PR #1, opened 2022-09-26
Files api: add "receive", "receive content" and "delete"

gehrisandro:
Hi @nunomaduro 

I had a little time and extended the files api a bit.

Was not 100% sure how we should approach the content receiving as this will not follow the REST convention to return an array. Instead it returns just a string.
In order to keep the type definitions as tight as possible I've added a new method (`requestContent()`) on the Transport. Hope this is fine for you. Otherwise just let me know.


nunomaduro:
Very impressive pull request. Good job! Feel free to continue, and implement other methods. ✅

PR #2, opened 2022-09-27
Files api: add "upload"

gehrisandro:
Hi @nunomaduro 

A PR to add the missing file upload.

The "Content-Type" header creation is a bit ugly but I had no better idea while still using your initial concept with the PSR7-Request creation.


nunomaduro:
Thank you!!

Issue #4, opened 2022-09-29
fully typed responses and requests


gehrisandro:
Hi @nunomaduro 

First of all, thank you for reviewing and merging my previous PRs so quickly! 👍 
It's a pleasure to help you with this package. Learned already a lot about (Open)AI and even more from your way how to build a clean package.

I am a huge fan of using fully typed responses and requests. Therefore I gave it a try with the moderations endpoint to see how it could work.

What I ended up with is the following:

```php
$client = OpenAI::client('TOKEN');

$request = new ModerationCreateRequest(input: 'I want to kill them.', model: ModerationModel::TextModerationLatest);

$response = $client->moderations()->create($request);

dump($response->id); // modr-5vvCuUd3dRjgIumIZIu0yBepv5qwL
dump($response->model); // text-moderation-003
dump($response->results[0]->flagged); // true
dump($response->results[0]->categories[0]->toArray()); // ["category" => "hate", "violated" => true, "score" => 0.40681719779968 ]
```

In my opinion this gives the developers the better UX than plain arrays.

More or less I took the approach Steve McDougall described here: https://laravel-news.com/working-with-data-in-api-integrations

I also implemented request factories to give the user various options how to create the request instance:

```php
// create the request directly
$request = new ModerationCreateRequest(
    input: 'I want to kill them.',
    model: ModerationModel::TextModerationLatest,
);

// pass an array to a factory instance
$request = (new ModerationCreateRequestFactory)->make([
    'input' => 'I want to kill them.',
    'model' => ModerationModel::TextModerationLatest,
]);

// pass an array to a static factory method
$request = ModerationCreateRequestFactory::new([
    'input' => 'I want to kill them.',
    'model' => ModerationModel::TextModerationLatest,
]);
```

If you want to have a look, I pushed the POC here: https://github.com/gehrisandro/openai-php-client/tree/poc-strong-typed-requests-and-responses

nunomaduro:
I am in favour of typing responses. Can you make a regular pull request and we work from there?

PR #5, opened 2022-09-29
[0.x] Adds `Moderations::create` typed response

gehrisandro:
There we go.

see #4


nunomaduro:
@gehrisandro It's possible to focus this pull request in **only on typed responses**, and on a single resource method?

gehrisandro:
@nunomaduro I narrowed the pull request down to focus only on the typed response as you suggested.

In my opinion the typed response is very easy to understand and use.
But still there are some points I wasn't sure what the best approach would be:

- Is it a good idea to rearrange the data returned from the API or should the structure represented in exactly the same way? For the moderation endpoint, I rearranged the resulting categories data  from two arrays having the same keys to a single array with category objects, which feels for me more natural.

- The response objects are namespaced under "DataObjects" and not just simple "Responses". In my opinion it expresses a bit better, that it isn't a 100% exact copy of what the API returned. (only if you agree to rearrange data as mentioned above)

- Naming of the nested data objects: Moderation\ModerationCategory vs. Moderation\Category
Probably it's an endless discussion which approach is better. In my opinion most important is to be consistent within the project. What do you prefer?

- To create the data objects from the array I've created dedicated factory classes. Maybe it's a bit over engineered but it keeps the API call method every clean as it can simply call the factory with the data returned from the API.

- Last but not least. I had to add a `@phpstan-ignore-line` in ModerationFactory::make() as I wasn't able to find a proper type declaration which doesn't lead to an error.

I'm looking forward to hearing your opinion.

src/DataObjects/Moderation/Moderation.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace OpenAI\DataObjects\Moderation;
```
nunomaduro, line 3: This should be `OpenAI\Responses\Moderations;`.

src/DataObjects/Moderation/Moderation.php
```diff
@@ -0,0 +1,30 @@
...
+
+use OpenAI\Contracts\DataObject;
+
+final class Moderation implements DataObject
```
nunomaduro, line 7: This class should be name `CreateResponse` and it should implement the `Response` interface.

src/DataObjects/Moderation/ModerationCategory.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace OpenAI\DataObjects\Moderation;
```
nunomaduro, line 3: OpenAI\Responses\Moderations;

src/DataObjects/Moderation/ModerationCategory.php
```diff
@@ -0,0 +1,28 @@
...
+
+use OpenAI\Contracts\DataObject;
+use OpenAI\Enums\Moderation\Category;
+
+final class ModerationCategory implements DataObject
```
nunomaduro, line 8: This one should be named CreateResponseModerationCategory, etc.

nunomaduro:
@gehrisandro Left a few comments on the way I would probably tackled this.

gehrisandro:
Thank you for our feedback.

I renamed classes / namespaces in the way to suggested, updated and cleaned update the tests.

nunomaduro:
Adjusting things locally.

PR #7, opened 2022-10-04
 [0.x] Adds typed responses for more resources

gehrisandro:


gehrisandro:
While implementing more typed responses I've produced more and more duplicated / similar code.
For example `OpenAI\Responses\Files\DeleteResponse` and `OpenAI\Responses\Models\DeleteResponse` are more or less identical. Same for `OpenAI\Responses\Files\CreateResponse` and `OpenAI\Responses\Files\ReceiveResponse`

I will wait with a refactor until the last resource (FileTunes) is implemented because then we will have a complete picture of all responses which are more or less redundant.

TODO:

- [ ] Implement FileTunes resource
- [ ] Refactor redundant response implementations

nunomaduro:
@gehrisandro I think is fine "duplicated" code on value objects. So, going to wait for the `FileTunes` before validating the full thing.

gehrisandro:
@nunomaduro `FileTunes` is already done.

PR #9, opened 2022-10-27
feat: add two more specific exception classes

gehrisandro:
Added a more fine granular exception handling for all so far known exceptions. But I wasn't able to find a documentation about all the errors the API possibly returns.

If the exception type is unknown it throws the existing `ErrorException`


gehrisandro:
As long there is no official documentation on the possible errors I do not think that it is worth to have more custom exception classes.

PR #13, opened 2022-10-29
wip: create request object for a completion create request

gehrisandro:
Tinkered around a bit creating a request object similar to the existing response objects to provide the users a bit more guidance when creating an API request.

Curious to hear what do you think @nunomaduro 

There are four possibilities to create a request:

### Create a request through the constructor
```php
$request = new CreateCompletionRequest(
    model: 'text-davinci-002',
    prompt: 'PHP is ',
    maxTokens: 10,
);

$response = $client->completions()->create($request);
```

### Complete the requests with only the required parameter and than set the properties individually
```php
$request = new CreateCompletionRequest('text-davinci-002');
$request->prompt = 'PHP is';
$request->maxTokens = 10;

$response = $client->completions()->create($request);
```

### Create a request from the array (same format the API expects)
```php
$request = CreateCompletionRequest::from([
    'model' => 'text-davinci-002',
    'prompt' => 'PHP is',
    'max_tokens' => 10,
]);

$response = $client->completions()->create($request);
```

### Pass the array directly as it was before (behind the scene it creates a request using the `from()` method)
```php
$response = $client->completions()->create([
    'model' => 'text-davinci-002',
    'prompt' => 'PHP is',
    'max_tokens' => 10,
]);
```

Except from giving the user more guidance it also introduces some more strictness on the data send to the API. It's not anymore possible to send unexpected fields or pass for example a string to a field that expects a boolean.

BTW: Didn't updated the type hinting and tests yet.


gehrisandro:
Hi @nunomaduro 

The request object for the completion create response is now implemented completely.

What do you think? Is there anything you would change / improve?

If you agree I would continue to integrate request objects for all the other endpoints and release them continuously  as they are not breaking the current implementation (passing a raw array is still valid).

nunomaduro:
Putting this on hold.

PR #22, opened 2022-12-22
Use latest davinci text model

mpociot:
This PR updates the README to use the latest text-davinci model, so that examples can be easily copy/pasted from the readme.


nunomaduro:
Thank you!

PR #25, opened 2022-12-28
Naming fixed of ApiKey and variable names

filips-dev:


tests/ValueObjects/ApiToken.php
```diff
@@ -2,10 +2,10 @@
 
 // Generate test for API token value object...
```
nunomaduro, line 3: @filips-dev Can you adjust this comment and adjust the file name of this test?

nunomaduro (changes requested):

tests/ValueObjects/ApiToken.php
filips-dev: Resolved

Issue #23, opened 2022-12-28
Value object and arguments with the name `ApiToken|$apiToken` should be `ApiKey|$apiKey`


nunomaduro:
We've just noticed that, within this project, we are using the terminology `ApiToken` and `$apiToken`, where in fact, it should `ApiKey` and `$apiKey`, as mentioned in multiple places in OpenAI documention.

filips-dev:
@nunomaduro Seeing as I'm new to Open source how can I push my branch and make a PR?

nunomaduro:
Fixed here: https://github.com/openai-php/client/pull/25.

PR #26, opened 2022-12-28
Renamed ApiToken to ApiKey

iamharis010:
This PR is introduced to replace the term ApiToken|apiToken to ApiKey|apiKey as mentioned by Nuno in PR Request.

https://github.com/openai-php/client/issues/23

Also ran the test for cross check. All of them are passed.

image


nunomaduro:
Thank you. But this issue is already addressed: https://github.com/openai-php/client/pull/25.

Issue #30, opened 2022-12-29
Make base URI configurable


LeoVie:
Hi,

what do you think about making the base URI configurable? At the moment, the base URI is hardcoded to `https://api.openai.com/v1`.

Making it configurable would make end-to-end testing of applications using the OpenAI client easier, as one could use an url to a mocking server in the test environment.

I would implement this as a non-breaking change via the following steps:
1. Extract interface from `OpenAI\ValueObjects\Transporter\BaseUri`
2. Change parameter type of `$baseUri` in `OpenAI\ValueObjects\Transporter\Payload::toRequest` to extracted interface
3. Add optional parameter `BaseUriInterface $baseUri = null` to `OpenAI::client`
4. Modify `OpenAI::client`, so that it handles the default value for `$baseUri` like that:
```php
public static function client(string $apiToken, string $organization = null, BaseUriInterface $baseUri = null): Client
{
    ...
    $baseUri = $baseUri ?? BaseUri::from('api.openai.com/v1');
    ...
}
```

What do you think about that? If you don't object, I would implement that.

nunomaduro:
Just add protocol, via the `BaseUri`, if it's missing from the given string.

LeoVie:
Okay, I will prepare a PR later that week :)

LeoVie:
Hmm, I guess, doing it via

```php
OpenAI::client('sk-...')
    ->withBaseUri('http://api.foo-base.com/v1');
```

is not that trivial, as the `Client` created via `OpenAI::client` does not know the base URI directly, but via `HttpTransporter`. 

I've created a PR, that allows usage in the form
```php
OpenAI::clientWithBaseUri('sk-...', 'http://api.foo-base.com/v1');
```

I'm open to opinions on this :)

gehrisandro:
This can now be achieved by using the newly introduced factory:

```php
$client = OpenAI::factory()
    ->withApiKey('<your-api-key>')
    ->withBaseUri('openai.example.com/v1')
    ->make();
```

PR #31, opened 2022-12-29
#30 Add method to create client with custom base URI

LeoVie:
This solves #30. Please also see my comment here https://github.com/openai-php/client/issues/30#issuecomment-1367486440

`BaseUri::toString` will now check, if the given base URI starts with a protocol. If not, it adds `https://` as protocol as prefix to the URI.

Also, the method will now check, if the given base URI ends with a slash and only append a trailing slash to it, if not.

That can be used, so users can create clients with a custom base URI, that is not required to use `https` as protocol, e.g.

```php
$client = OpenAI::clientWithBaseUri('sk-...', 'http://mock-api/v1');
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,15 @@ $result = $client->completions()->create([
 echo $result['choices'][0]['text']; // an open-source, widely-used, server-side scripting language.
 ```
 
+You can also instantiate the client with an organization or with a custom base api URI:
+
+```php
+$client = OpenAI::factory('YOUR_API_KEY')
+    ->withOrganization('YOUR_ORGANIZATION')
+    ->withBaseUri('CUSTOM_BASE_URI')
+    ->make();
+```
+
 ## Usage
 
 ### `Models` Resource
--- a/src/Factory/ClientFactory.php
+++ b/src/Factory/ClientFactory.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenAI\Factory;
+
+use GuzzleHttp\Client as GuzzleClient;
+use OpenAI\Client;
+use OpenAI\Transporters\HttpTransporter;
+use OpenAI\ValueObjects\ApiKey;
+use OpenAI\ValueObjects\Transporter\BaseUri;
+use OpenAI\ValueObjects\Transporter\Headers;
+
+final class ClientFactory
+{
+    public function __construct(
+        private readonly string $apiKey
+    ) {
+    }
+
+    private string $baseUri = 'api.openai.com/v1';
+
+    private ?string $organization = null;
+
+    public function make(): Client
+    {
+        $apiKey = ApiKey::from($this->apiKey);
+
+        $baseUri = BaseUri::from($this->baseUri);
+
+        $headers = Headers::withAuthorization($apiKey);
+
+        if ($this->organization !== null) {
+            $headers = $headers->withOrganization($this->organization);
+        }
+
+        $client = new GuzzleClient();
+
+        $transporter = new HttpTransporter($client, $baseUri, $headers);
+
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -34,6 +34,12 @@ $result = $client->completions()->create([
 echo $result['choices'][0]['text']; // an open-source, widely-used, server-side scripting language.
 ```
 
+You can also instantiate the client with a custom base api URI:
+
+```php
+$client = OpenAI::clientWithBaseUri('YOUR_API_KEY', 'CUSTOM_BASE_URI');
```
nunomaduro, line 40: Actually, after giving some thoughts into this. I think you should do something different:

```php
OpenAI::factory($apiKey) // returns factory
    ->withBaseUri('string') // returns factory
    ->make(); // makes client.
```

nunomaduro (changes requested):

README.md
LeoVie: I've implemented it that way. Please take a look, again :)

src/Factory/ClientFactory.php
```diff
@@ -0,0 +1,57 @@
...
+namespace OpenAI\Factory;
+
+use GuzzleHttp\Client as GuzzleClient;
+use OpenAI\Client;
+use OpenAI\Transporters\HttpTransporter;
+use OpenAI\ValueObjects\ApiKey;
+use OpenAI\ValueObjects\Transporter\BaseUri;
+use OpenAI\ValueObjects\Transporter\Headers;
+
+final class ClientFactory
```
nunomaduro, line 14: Can you add some PHP annotations - like we on the rest of the code to this class methods / properties.

tests/Factory/ClientFactory.php
```diff
@@ -0,0 +1,57 @@
...
+
+    $expected = new Client(
+        new HttpTransporter(
+            new GuzzleClient(),
+            BaseUri::from('http://api.custom-domain.com/v1'),
+            Headers::withAuthorization(ApiKey::from('foo'))
+        )
+    );
+
+    expect($openAI)->toEqual($expected);
```
nunomaduro, line 39: What exactly this `toEqual` is testing? You may have to use reflection - to see if the client was created with the correct information instead.

nunomaduro (changes requested):
The pull request is perfect! Left a few comments for some final adjustments.

tests/Factory/ClientFactory.php
LeoVie: `toEqual` tests, if the correct base URI / organization was injected into the `HttpTransporter`.
The test fails, when that's not the case (e.g. when `withBaseUri` does nothing).

See also https://3v4l.org/ZobO3#v8.1.13 -> Comparing objects with `==` compares the primitive values in the object (in contrast to `===`).

Or what do you mean?

src/Factory/ClientFactory.php
LeoVie: tbh, I don't know, what PHP annotations could bring any value here. Do you mean comments explaining the properties?

tests/Factory/ClientFactory.php
nunomaduro: lol, TIL.

src/Factory/ClientFactory.php
LeoVie: Do you need any further information from me, @nunomaduro? :)

LeoVie:
Hi, I don't want to be impatient, but do you need any more information from me? It's also fine, if you have doubts about this changes, would be just good to know :)

gehrisandro:
An extended factory has been added in v0.4.0.

@LeoVie Thank you very much for your help and patience.

PR #33, opened 2023-01-02
Fixed broken getting of body as a string

GrahamCampbell:
`->getContents()` only reads the stream from wherever the current position is - it does not rewind it. This can be hugely problematic if, for example, a logging middleware is attached to the client, which leaves the pointer not at the start of the stream/


nunomaduro:
@GrahamCampbell Any chance of having a test on this?

PR #48, opened 2023-02-23
Missing namespace causes OpenAI class to be unusable

jamesrweb:
Since the namespace was not included, importing the class to use the static `client` method is impossible. It also creates a situation where the documentation is wrong since it uses this class and static method combination. We should check if anywhere else has such issues and release this ASAP in a fix version.

### DIFF
--- a/src/OpenAI.php
+++ b/src/OpenAI.php
@@ -2,6 +2,8 @@
 
 declare(strict_types=1);
 
+namespace OpenAI;
+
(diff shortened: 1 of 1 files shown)


nunomaduro:
The class gets included via Composer: https://github.com/openai-php/client/blob/9da3c248d3d0da3d26ff047168479a8becbc07d1/composer.json#L31.

Issue #51, opened 2023-02-26
Error: NULL finish_reason for completions


afedukov:
Sometimes getting error:
```
OpenAI\Responses\Completions\CreateResponseChoice::__construct(): Argument #4 ($finishReason) must be of type string, null given, called in /var/www/vendor/openai-php/client/src/Responses/Completions/CreateResponseChoice.php on line 24
```
$attributes from `\OpenAI\Responses\Completions\CreateResponseChoice::from`:
```
array(4) {
  ["text"]=>
  string(972) " ... somthing here ... "
  ["index"]=>
  int(0)
  ["logprobs"]=>
  NULL
  ["finish_reason"]=>
  NULL
}
```

careybaird:
Duplicate of #50

nunomaduro:
Fixed on `v0.3.2`.

Issue #50, opened 2023-02-25
TypeError when constructing OpenAI response choice object with null finish reason.


ManukMinasyan:
```php
TypeError: OpenAI\Responses\Completions\CreateResponseChoice::__construct(): Argument #4 ($finishReason) must be of type string, null given, called in /vendor/openai-php/client/src/Responses/Completions/CreateResponseChoice.php](*/vendor/openai-php/client/src/Responses/Completions/CreateResponseChoice.php) on line 26
  File "/app/Actions/Document/CreateNewContent.php", line 64, in App\Actions\Document\CreateNewContent::complete
    $result = OpenAI::completions()->create($parameters);
  File "/app/Actions/Document/CreateNewContent.php", line 39, in App\Actions\Document\CreateNewContent::App\Actions\Document\{closure}
    return $this->complete($template, $document, $data);
  File "/app/Actions/Document/CreateNewContent.php", line 42, in App\Actions\Document\CreateNewContent::create
    });
  File "/app/Http/Controllers/App/DocumentController.php", line 182, in App\Http\Controllers\App\DocumentController::writeForMe
    $choices = $contentCreator->create($template, $document, $data);
  File "/public/index.php", line 52
    $request = Request::capture()
...
(77 additional frame(s) were not displayed)
```

NumairAwan:
Getting the same error. Any solution?

careybaird:
Started seeing this also. Seems to be a recent breaking change in the OpenAI API.

The documentation doesn't say much about valid values for finish_reason, so I suggest we should treat it as potentially null.

I can do a PR maybe later today, will check in here in a few hours.

careybaird:
Pull request made: https://github.com/openai-php/client/pull/52

careybaird:
If anyone else wants to apply this fix without waiting for the new release (and you're using composer), change the package in your composer file like this and run `composer update`

`"openai-php/client": "dev-main as 0.3.1",`

Note that we are aliasing it as 0.3.1 so it will still work with other packages that use it as a dependency (e.g. openai-php/laravel)

nunomaduro:
Fixed on `v0.3.2`.

PR #52, opened 2023-02-26
Make finishReason nullable #51

careybaird:
A recent change to the OpenAI API means that finish_reason is sometimes returned as NULL and causes an exception.

This change allows finishReason to be nullable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ foreach ($response->choices as $result) {
     $result->text; // '\n\nThis is a test'
     $result->index; // 0
     $result->logprobs; // null
-    $result->finishReason; // 'length'
+    $result->finishReason; // 'length' or null
 }
 
 $response->usage->promptTokens; // 5,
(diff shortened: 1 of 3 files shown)


JonPurvis:
Thanks for this @careybaird, was just about to report it as an issue and then saw your PR, I'd say I get this 60-70% of the time, hopefully this will be merged soon 🥳

careybaird:
tagging @nunomaduro as I believe this issue is urgent, it affects everyone

janverhulst:
Same issue here, really urgent for us as well, since we cannot continue developing our software without this.

williamjulianvicary:
Came here to make the same PR - thanks for pushing the change - agreed this should be a priority merge, affecting a lot of requests.

sgasser:
@nunomaduro can you tag a new release with this fix? Thanks

careybaird:
If you want a workaround to use this now, see this: https://github.com/openai-php/client/issues/50#issuecomment-1447704749

nunomaduro:
Fixed on `v0.3.2`.

Issue #45, opened 2023-02-16
An issue on Fine Tune List API: when the status_details of a RetrieveResponseFile is an exception message (string)


rodrigoisabelo:
Hi, I encountered an issue with retrieving the list of fine tunes. 
1. I uploaded a JSONL file
2. Used the file to create a new fine-tune
3. When I tried to retrieve the list of fine tunes.
4. It gave me this error.

I suspect this is because the status details show that the file I uploaded was invalid. However, this is not really an issue with the package. But it would be great if the package could also handle this scenario.

I hope this can be resolved soon. Thanks!

OpenAI\Responses\FineTunes\RetrieveResponseFile::__construct(): Argument #8 ($statusDetails) must be of type ?array, string given, called in /var/www/html/vendor/openai-php/client/src/Responses/FineTunes/RetrieveResponseFile.php on line 50

The exception upon checking:
![image](https://user-images.githubusercontent.com/100250620/219321917-fcbc8f4d-93ee-4aaf-aae4-8e8e5e03aa42.png)

The stack trace I received:
![image](https://user-images.githubusercontent.com/100250620/219321988-80218930-59b2-4de7-8277-74acda0b9de2.png)

nunomaduro:
Can you make a pull request?

shutupflanders:
This is still an issue, can it be reopened?

`OpenAI\Responses\FineTunes\RetrieveResponseFile::__construct(): Argument #8 ($statusDetails) must be of type ?array, string given, called in /app/vendor/openai-php/client/src/Responses/FineTunes/RetrieveResponseFile.php on line 50`

rodrigoisabelo:
Hi @shutupflanders, it's good to know there are others who have experienced this issue, can you provide us again the step-by-step guide on how to replicate it?  Wasn't able to give @gehrisandro since I already refactored my code. So that he can help us and prevent this to happen to other developers as well. Thanks!

Will re-open this issue.

shutupflanders:
I uploaded a massive JSONL document with parsing errors in it (unescaped quote) by mistake, saw the error above, and got the following when hitting the `/v1/fine-tunes` endpoint manually:

`"status_details": "Invalid file format. Example 1196 cannot be parsed. Error: line contains invalid json: Expecting ',' delimiter: line 1 column 65 (char 64) (line 1196)"`

rodrigoisabelo:
[...]

PR #55, opened 2023-02-28
Changed Files/RetrieveResponse status_details parameters to mixed

rodrigoisabelo:
Hi, I am not really an expert in contributing or developing open-source projects, but let me know if you want specific edits.

This solves issue #45

### DIFF
--- a/src/Responses/Files/RetrieveResponse.php
+++ b/src/Responses/Files/RetrieveResponse.php
@@ -28,7 +28,7 @@ final class RetrieveResponse implements Response
         public readonly string $filename,
         public readonly string $purpose,
         public readonly string $status,
-        public readonly ?array $statusDetails,
+        public readonly ?mixed $statusDetails,
     ) {
     }
 
--- a/src/Responses/FineTunes/RetrieveResponseFile.php
+++ b/src/Responses/FineTunes/RetrieveResponseFile.php
@@ -28,7 +28,7 @@ final class RetrieveResponseFile implements Response
         public readonly string $filename,
         public readonly string $purpose,
         public readonly string $status,
-        public readonly ?array $statusDetails,
+        public readonly ?mixed $statusDetails,
     ) {
     }
 


src/Responses/Files/RetrieveResponse.php
```diff
@@ -28,7 +28,7 @@ private function __construct(
         public readonly string $filename,
         public readonly string $purpose,
         public readonly string $status,
-        public readonly ?array $statusDetails,
+        public readonly ?mixed $statusDetails,
```
nunomaduro, line 31: What's the PHP's type we can have here?

nunomaduro (changes requested):

src/Responses/Files/RetrieveResponse.php
rodrigoisabelo: Hi @nunomaduro , The data that I currently have as a basis for this is not yet clear, if the training files were uploaded successfully, I only got `null` for `statusDetails`, but the only time it has value is when there's an error and the data type is `string` for the exception message from OpenAI.

Should I change it to `public readonly ?string $statusDetails` instead? I'm asking you this since you far much more knowledgeable and I want to know what works best for you.

gehrisandro: @rodrigoisabelo I tried to figure out what the correct type should be, but the API specification is not very clear what we can expect here: https://github.com/openai/openai-openapi/blob/master/openapi.yaml#L3042

The spec says it is an `object` (without further clarification) or `null`. Therefore the current implementation with `?array` seems to be reasonable.

Furthermore I tried to reproduce your issue where it returns a string instead of null (or an object) but I was not able to do so as I always got `null`.

Can you provide us a way to reproduce the issue?

rodrigoisabelo: Try replacing the extension of this file into JSONL and upload it on OpenAI files API. It might take some time to produce an error. But from what have experienced the status details were an exception message from them.

[invalid_training_data.txt](https://github.com/openai-php/client/files/10866294/invalid_training_data.txt)

gehrisandro: @rodrigoisabelo I am not able to upload this file to OpenAI. Already during the upload I've got an error response:

```
OpenAI\Exceptions\ErrorException {
  #message: "Expected file to have JSONL format, where every line is a JSON dictionary. Line 2 is not a dictionary (HINT: line starts with: "{"pr...")."
  #code: 0
  #file: "/Users/sandro/Code/openai-php-client/src/Transporters/HttpTransporter.php"
  #line: 61
  ...
```

Probably you run into some strange malfunctioning on the API side. As long it is not possible to reproduce, I don't think it's worth to fix it on client side especially the change would go against their API documentation.

rodrigoisabelo: It's fine @gehrisandro , it's not that we are always uploading incorrect files though. I guess I just missed some things on my end as well. Thanks for looking into it. I'll be closing this pull request now.

Issue #64, opened 2023-03-03
Symfony Bundle


GromNaN:
Hi, thank you for this library that is very easy to use.

I created a Symfony bundle by copying things from the Laravel integration. You can find it here: https://github.com/GromNaN/openai-symfony (work in progress).

What do you think of moving this project into the `openai-php` organisation? It would be good for this project to provide an integration with a 2nd major framework. The bundle is not published on Packagist yet, so that it would be a clean start.

nunomaduro:
Excellent idea. Just created the repository, and added you as maintainer. Can you make sure the project, structure, coding style, readme, (everything really), looks exactly like the client / laravel integration? Thanks! 

- https://github.com/openai-php/symfony.
- https://github.com/openai-php/symfony/invitations.

GromNaN:
Excellent. I'm going to do this right.

nunomaduro:
Let me know when is ready, and i will review it. ✔️

GromNaN:
Ready. You can submit on Packagist whenever you want to.

There is a [phpstan error](https://github.com/openai-php/symfony/actions/runs/4327840200) that I'll dig with the Symfony team.

nunomaduro:
Done. So, now I guess we have the HTTP client issue + the tests not passing issue, that's correct?

GromNaN:
Thank you for the publication on [packagist](https://packagist.org/packages/openai-php/symfony).
Tests are green after https://github.com/openai-php/symfony/commit/c7d4a74322f7afd8f83f708ef99bf9b8bb579be0 (and the bundle is working again).

GromNaN:
Closing as the bundle have been tagged. Thanks.

PR #65, opened 2023-03-03
Allows to change the HTTP client in the factory

GromNaN:
[Symfony integration](https://github.com/openai-php/client/issues/64) would be complete if the `symfony/http-client` could be used. Instead of duplicating the `OpenAI::client()` factory, I think it is more simple to allow a custom instance of `Psr\Http\Client\ClientInterface`. The [Symfony PSR-18 client](https://symfony.com/doc/current/http_client.html#psr-18-and-psr-17) would be used.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,9 +11,10 @@
     ],
     "require": {
         "php": "^8.1.0",
-        "guzzlehttp/guzzle": "^7.5.0"
+        "psr/http-client-implementation": "1.0"
     },
     "require-dev": {
+        "guzzlehttp/guzzle": "^7.5.0",
         "laravel/pint": "^1.6.0",
         "nunomaduro/collision": "^7.0.5",
         "pestphp/pest": "^2.0.0",
@@ -22,6 +23,9 @@
         "phpstan/phpstan": "^1.10.3",
         "rector/rector": "^0.14.8",
         "symfony/var-dumper": "^6.2.7"
+    },
+    "suggest": {
+        "guzzlehttp/guzzle": "^7.5.0"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,8 @@
     ],
     "require": {
         "php": "^8.1.0",
+        "guzzlehttp/psr7": "^2.4.4",
+        "php-http/discovery": "^1.14",
         "psr/http-client-implementation": "1.0"
     },
     "require-dev": {
@@ -23,9 +25,6 @@
         "phpstan/phpstan": "^1.10.3",
         "rector/rector": "^0.14.8",
         "symfony/var-dumper": "^6.2.7"
-    },
-    "suggest": {
-        "guzzlehttp/guzzle": "^7.5.0"
     },
     "autoload": {
         "psr-4": {
@@ -46,7 +45,8 @@
         "sort-packages": true,
         "preferred-install": "dist",
         "allow-plugins": {
-            "pestphp/pest-plugin": true
+            "pestphp/pest-plugin": true,
+            "php-http/discovery": false
         }
     },
     "scripts": {
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "php": "^8.1.0",
         "guzzlehttp/psr7": "^2.4.4",
         "php-http/discovery": "^1.14",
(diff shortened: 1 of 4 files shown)


tests/OpenAI.php
```diff
@@ -13,3 +16,15 @@
 
     expect($openAI)->toBeInstanceOf(Client::class);
 });
+
+it('accepts a custom http client', function () {
+    $client = new class implements ClientInterface {
```
GromNaN, line 21: I don't know this test framework. I'm not sure how to make a mock.

composer.json
```diff
@@ -11,9 +11,10 @@
     ],
     "require": {
         "php": "^8.1.0",
-        "guzzlehttp/guzzle": "^7.5.0"
+        "psr/http-client-implementation": "1.0"
```
GromNaN, line 14: I propose to require `psr/http-client-implementation` instead of a specific implementation. So that it is not necessary to exclude guzzle from projects that want to use an other implementation.

For the Laravel package, guzzle can be explicitly required to keep a smooth installation path.

nicolas-grekas: this should also require `"php-http/discovery": "^1.15"` so that such an implementation is installed automatically (+ the plugin should be disabled as such since we don't need it in for testing this:
```
    "config": {
        "allow-plugins": {
            "php-http/discovery": false
        }
    },
```

composer.json
```diff
@@ -23,6 +24,9 @@
         "rector/rector": "^0.14.8",
         "symfony/var-dumper": "^6.2.7"
     },
+    "suggest": {
+        "guzzlehttp/guzzle": "^7.5.0"
```
nicolas-grekas, line 28: not needed if php-http/discovery is added

src/OpenAI.php
```diff
@@ -26,7 +27,12 @@ public static function client(string $apiKey, string $organization = null): Clie
             $headers = $headers->withOrganization($organization);
         }
 
-        $client = new GuzzleClient();
```
nicolas-grekas, line 29: `$client ??= HttpClientDiscovery::find()` instead

composer.json
GromNaN: If guzzle is required by the laravel package, can we skip the composer plugin?

nicolas-grekas: then the plugin will do nothing :shrug: if you fear composer asking for enabling the plugin, it's already whitelisted by both laravel and symfony, and it's anyway still better than composer failing to install because of a missing implementation package.
But you might prefer going with `^1.14` so that this doesn't conflict with sentry when it's installed.

composer.json
```diff
@@ -11,9 +11,12 @@
     ],
     "require": {
         "php": "^8.1.0",
-        "guzzlehttp/guzzle": "^7.5.0"
+        "guzzlehttp/psr7": "^2.4.4",
+        "php-http/discovery": "^1.14",
+        "psr/http-client-implementation": "1.0"
```
nicolas-grekas, line 16: `^1.0` btw on this line

tests/OpenAI.php
```diff
@@ -13,3 +16,17 @@
 
     expect($openAI)->toBeInstanceOf(Client::class);
 });
+
+it('accepts a custom http client', function () {
+    $client = new class implements ClientInterface
+    {
```
nicolas-grekas, line 22: should be on the previous line, isn't it?

tests/OpenAI.php
```diff
@@ -13,3 +16,17 @@
...
+        public function sendRequest(RequestInterface $request): ResponseInterface
+        {
+            throw new \LogicException('Not implemented');
+        }
+    };
+
+    $openAI = OpenAI::client('foo', client: $client);
```
nicolas-grekas, line 29: I feel like the test should either ensure that the passed client is actually used when consuming the API, or that putting `client: null` would provide the same "safety" here

GromNaN (commented):
Thanks for your review @nicolas-grekas. I hope maintainers will like this framework-agnostic implementation.

nicolas-grekas (approved):
LGTM on my side thanks.

gehrisandro:
Completed in https://github.com/openai-php/client/pull/75

Thanks to @nicolas-grekas @GromNaN for your help and your valuable feedback!

Issue #46, opened 2023-02-16
change Endpoint to Azure


timmoh:
Since the open AI API is available in azure, is there a possibility to change to endpoint to azure.
or will there be a plan to add this feature

gehrisandro:
Hi @timmoh 

Thank you for raising this issue.

I took a closer look into the Azure documentation and it looks like it is not enough to just change the endpoint url. It additionally requires an `api-version` query parameter.
See: https://learn.microsoft.com/en-us/azure/cognitive-services/openai/reference#completions

And the endpoint url itself is not static too as it contains your personal `resource-name` and `deployment-id`

Are you using Azure and can you confirm my findings?

duellsy:
Just jumping in here, I can confirm this. The  URL structure is for example:

`{openai_api_base}/openai/deployments/{deployment_name}/completions?api-version={openai_api_version}`

gehrisandro:
Hi @duellsy 

Thank you for the confirmation.

I think the upcoming change to the package would help here: https://github.com/openai-php/client/pull/75

The only thing we would have to add is a new method to the factory where arbitrary values can be passed to be used as query parameters:
```php
->withQueryParams(['api-version' => '<version>'])
```

duellsy:
Neat 🙌

I couldn't see how that would end up with the correct full endpoint as above though?
Unless you were meaning that it's currently a big step in that direction, in which case my apologies.

gehrisandro:
Sorry, if it wasn't clear 😅  Here is the full example:

```php
$client = OpenAI::factory()
    ->withBaseUrl('https://{your-resource-name}.openai.azure.com/openai/deployments/{deployment-id}')
    ->withHttpHeader('api-key', '{your-api-key}')
    ->withQueryParams(['api-version' => '{version}'])
    ->make();
```

duellsy:
Ha! Beat me to it, I was just about to say the entire URL up to `/completions` could be used as the base URL ✌️

gehrisandro:
This should now work using the newly introduced factory:

```php
$client = OpenAI::factory()
    ->withApiKey('<your-api-key>')
    ->withBaseUri('{your-resource-name}.openai.azure.com/openai/deployments/{deployment-id}')
    ->withHttpHeader('api-key', '{your-api-key}')
    ->withQueryParam('api-version', '{version}')
    ->make();
```

@duellsy Could you please verify this for us?

duellsy:
Nice one. It works well, a few notes:

[...]

Issue #39, opened 2023-01-28
Add Timeout Param


Justintime50:
The official Python library allows a timeout to be set on requests. It would be really helpful for production applications to be able to set up a timeout on requests so we don't keep our web workers hanging if there is hiccups in connections or issues on the OpenAI side.

nunomaduro:
Can you make a pull request adding that feature?

Justintime50:
Hmmm, it doesn't even look like the [PSR7 lib](https://github.com/guzzle/psr7) that is used here to make HTTP calls supports timeouts at all if I'm looking through their code correctly? Would potentially require a wrapper around the requests to handle this since it's not supported native.

gehrisandro:
This can now be achieved by using the newly introduced factory:

```php
$client = OpenAI::factory()
    ->withHttpClient(new \GuzzleHttp\Client(['timeout'  => 2.0]))
    ->make();
```

PR #66, opened 2023-03-05
fix: set false GuzzleClient ssl laravel local developer

fabiovige:
During local development laravel throws ssl error. So I chose to add via service provider the verify = false option when instantiating the GuzzleClient that way it will not verify the ssl during local development.

### DIFF
--- a/src/OpenAI.php
+++ b/src/OpenAI.php
@@ -14,7 +14,7 @@ final class OpenAI
     /**
      * Creates a new Open AI Client with the given API token.
      */
-    public static function client(string $apiKey, string $organization = null): Client
+    public static function client(string $apiKey, string $organization = null, bool $verify = true): Client
     {
         $apiKey = ApiKey::from($apiKey);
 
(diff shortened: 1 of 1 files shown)


gehrisandro:
Hi @fabiovige 

Thank you for your suggestion.

Because we do have other requests to modify the client in one or the other way (#39 #46 #30) I think it's better to have a generic way to modify the client options instead of just adding an extra parameter for this specific purpose.

But we will take your requirement into consideration for the further development.
If you find time to look into the other issues, a more generic solution is very welcome.

fabiovige:
Claro,

Irei analisar com mais calma.

Obrigado pelo retorno.

gehrisandro:
@fabiovige Thank you.

And thanks to OpenAI is was able to understand your reply 😅 

```
Translate to english: "Claro,

Irei analisar com mais calma.

Obrigado pelo retorno."

Sure, I will take a closer look. Thank you for the feedback.
```

gehrisandro:
You can now use the newly introduce factory to customize the guzzle client:

```php
$client = OpenAI::factory()
    ->withApiKey('<your-api-key>')
    ->withHttpClient(new \GuzzleHttp\Client(['verify' => false]))
    ->make();
```

PR #71, opened 2023-03-11
add testing ability

gehrisandro:
This PR makes it easier to write tests in your application using this library.

This is related to https://github.com/openai-php/laravel/issues/23

@nunomaduro @mpociot  I have decided to create most of the testing logic within the client library itself and not in the Laravel wrapper. The main reason for this decision is to provide the same testing options to users which are using the client directly.
Additionally it should be fairly easy to provide the testing options in other wrappers too (@GromNaN)

The PR for the Laravel wrapper is here: https://github.com/openai-php/laravel/pull/27

There are multiple things to discuss:
- First of all: Do you agree to put most of the testing logic into the client library?
- I have add fixtures for all responses to make faking easier: It's not necessary to provide all response parameters, instead it's enough to provide the parameters the user is interested in. (see example below) Do you think that's too much?
- The response fixtures are at the moment classes having a constant containing the default response attributes. Maybe it could be changed to something different as it seems a bit too much to have classes.
- I have excluded the `OpenAI\Testing` namespace from `phpstan` because it's not really possible to add proper doc types as the return type could be possibly every available response.

Todos:
- [x] Add documentation

## A test using a fake response without providing all parameters
```php
it('returns the faked response', function () {
    $fake = new ClientFake([
        CreateResponse::fake([
            'choices' => [
                [
                    'text' => 'awesome!',
                ],
            ],
        ]),
    ]);

    $completions = $fake->completions()->create([
        'prompt' => 'PHP is ',
    ]);

    expect($completions['choices'][0])
        ->text->toBe('awesome!');
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## v0.4.0 (2023-03-17)
+### Changed
+- Removed dependency for `guzzlehttp/guzzle` and use PSR-18 client discovery instead ([#75](https://github.com/openai-php/client/pull/75))
+- Add Client factory which allows for a custom HTTP client
+- Client factory further accepts custom HTTP headers, query parameters and API URI
+
 ## v0.3.5 (2023-03-08)
 ### Fixed
 - `status_details` can be a string in file responses. Affects Files and FineTunes resources ([#68](https://github.com/openai-php/client/pull/68))
--- a/README.md
+++ b/README.md
@@ -767,6 +767,24 @@ $client->completions()->assertNotSent();
 $client->assertNothingSent();
 ```
 
+To write tests expecting the API request to fail you can provide a `Throwable` object as the response.
+
+```php
+$client = new ClientFake([
+    new \OpenAI\Exceptions\ErrorException([
+        'message' => 'The model `gpt-1` does not exist',
+        'type' => 'invalid_request_error',
+        'code' => null,
+    ])
+]);
+
+// the `ErrorException` will be thrown
+$completion = $client->completions()->create([
+    'model' => 'text-davinci-003',
+    'prompt' => 'PHP is ',
+]);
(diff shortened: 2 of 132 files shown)


mpociot:
@gehrisandro thank you very much for putting in the work for this PR.
I just tried the branch on What The Diff and it works great.

Internally, I personally don't think it's an issue that the testing implementation is inside the library.

There are two things that I noticed during implementing this:

> The fixture values used for default responses should be simpler.

While testing, I didn't specify all of the choices responses and I saw the default fixture values like "this is indeed a test". Those default values made me wonder for a second if I'm making some actual API requests, so I think it might be better to either leave them out entirely or use some more verbose defaults that clearly indicate that this is a fake response.

> Add the ability to test exceptions/errors

I have to add some tests to ensure that error handling is working correctly. This can have a couple of different reasons.
Either the API returns an internal error (GPT4 seems to do this), a rate limit error or something else.

I thought that I could just create a fake response with an error, but that didn't work as the `ErrorException` does not get thrown in the fake client:

```
CreateResponse::fake([
    'error' => [
        'message' => 'Something went wrong',
    ],
])
```

Other than that, this is perfect! :)

gehrisandro:
Hi @mpociot 

Thank you for your valuable feedback.

1) I will review the default values and make them more clearly.
Not providing any default would force the users to always provide a full API response to fake a response object and this is in my opinion less convenient because in most tests they will be only interested in a specific value.

2) To be honest I totally forgot about returning exceptions. I have to take a look how to implement this as you are going to create a fake response instance but in the client the exception handling takes place before the response object is instantiated.
For now I think we should allow to pass exceptions as fake responses.

@nunomaduro I will start adjusting to the stream responses. Hope to finish until Monday evening.

gehrisandro:
@mpociot Exception testing should work now:

```php
$client = new ClientFake([
    new \OpenAI\Exceptions\ErrorException([
        'message' => 'The model `gpt-1` does not exist',
        'type' => 'invalid_request_error',
        'code' => null,
    ])
]);

// the `ErrorException` will be thrown
$completion = $client->completions()->create([
    'prompt' => 'PHP is ',
]);
```

Issue #67, opened 2023-03-07
Add Proxy Support for HTTP Request


geekr-dev:
Can you add support for proxy setting for HTTP Client？for now,  classes in openai-php/client are all final，we can't extend them for customize

nunomaduro:
Can you elaborate on what do you mean by "proxy setting"?

seebyyu:
> Can you elaborate on what do you mean by "proxy setting"?

client [#73 ](https://github.com/openai-php/client/pull/73)
client-laravel [#25](https://github.com/openai-php/laravel/pull/25)

wdwlrz:
OpenAI.php
line:29
$client = new GuzzleClient(['verify'=>false,'proxy'=>'http://127.0.0.1:7890']);
do like this.
兄弟，clash挂个流量订阅，然后开始玩耍吧。

gehrisandro:
This can now be achieved by using the newly introduced factory:

```php
$client = OpenAI::factory()
    ->withApiKey('<your-api-key>')
    ->withHttpClient(new GuzzleClient(['verify'=>false,'proxy'=>'http://127.0.0.1:7890']))
    ->make();
```

PR #75, opened 2023-03-14
feat: add client factory

gehrisandro:
This adds the ability to configure the client instance via a new factory:

Excerpt from the updated README.
```php
$yourApiKey = getenv('YOUR_API_KEY');

$client = OpenAI::factory()
    ->withApiKey($yourApiKey)
    ->withOrganization('your-organization') // default: null
    ->withBaseUrl('openai.example.com/v1') // default: api.openai.com/v1
    ->withHttpHeader('X-My-Header', 'foo')
    ->withHttpClient(new \GuzzleHttp\Client([])) // default: HTTP client found using PSR-18 HTTP Client Discovery
    ->make();
```

This change is breaking as the "guzzlehttp/guzzle" has been removed. Existing projects without an explicitly installed HTTP client have to install the Guzzle Client manually:
```bash
composer require guzzlehttp/guzzle
``` 

With this change we can resolve various issues:
- #30 
- #31 
- #39 
- #65
- #67

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,9 @@ $client = OpenAI::factory()
     ->withApiKey($yourApiKey)
     ->withOrganization('your-organization') // default: null
     ->withBaseUrl('openai.example.com/v1') // default: api.openai.com/v1
-    ->withHttpHeader('X-My-Header', 'foo')
     ->withHttpClient(new \GuzzleHttp\Client([])) // default: HTTP client found using PSR-18 HTTP Client Discovery
+    ->withHttpHeader('X-My-Header', 'foo')
+    ->withQueryParam('my-param', 'bar')
     ->make();
 ```
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ $yourApiKey = getenv('YOUR_API_KEY');
 $client = OpenAI::factory()
     ->withApiKey($yourApiKey)
     ->withOrganization('your-organization') // default: null
-    ->withBaseUrl('openai.example.com/v1') // default: api.openai.com/v1
+    ->withBaseUri('openai.example.com/v1') // default: api.openai.com/v1
     ->withHttpClient(new \GuzzleHttp\Client([])) // default: HTTP client found using PSR-18 HTTP Client Discovery
     ->withHttpHeader('X-My-Header', 'foo')
     ->withQueryParam('my-param', 'bar')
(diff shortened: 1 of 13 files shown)


gehrisandro:
I am going to fix the failing checks later today...

gehrisandro:
@GromNan @nicolas-grekas What do you think? Does this what you need for the symfony integration?

README.md
```diff
@@ -24,6 +24,11 @@ First, install OpenAI via the [Composer](https://getcomposer.org/) package manag
 composer require openai-php/client
 ```
 
+If your project does not already include a PSR-18 client you have to install one:
```
GromNaN, line 27: You don't need this step if you use php-http/discovery.

GromNaN:
> What do you think? Does this what you need for the symfony integration?

The factory is useful pattern, but personally I prefer parameters (and named parameters), this is less methods to define and call.

README.md
gehrisandro: Thank you @GromNaN 

I've updated the composer.json to enable auto installation.
Updated the README accordingly.

Do you agree how it is written now?

seebyyu:
Hello, please adjust it as soon as possible [openai-php/laravel](https://github.com/openai-php/laravel)

huangdijia (approved):

composer.json
```diff
@@ -11,9 +11,15 @@
     ],
     "require": {
         "php": "^8.1.0",
-        "guzzlehttp/guzzle": "^7.5.0"
+        "php-http/discovery": "^1.15",
+        "php-http/multipart-stream-builder": "^1.2",
+        "psr/http-client-implementation": "^1.0.1",
```
nicolas-grekas, line 16: psr/http-factory-implementation should also be listed to benefit fromPsr17Factory

composer.json
```diff
@@ -11,9 +11,15 @@
...
+        "psr/http-client": "^1.0.1",
+        "psr/http-message": "^1.0.1"
     },
     "require-dev": {
+        "guzzlehttp/psr7": "^2.4.4",
```
nicolas-grekas, line 21: not needed since there is guzzle already

nicolas-grekas: Note that I'm not sure it's worth replacing this by php-http/multipart-stream-builder

composer.json
gehrisandro: thanks for the hint

composer.json
gehrisandro: It's not possible to remove it because we require here newer version.

nunomaduro (approved):

PR #85, opened 2023-03-23
Make ErrorException return types nullable

nurdism:
`ErrorException` will throw a `TypeError` when attempting to use `getErrorCode()` or `getErrorType()` if they are null, this RP makes those function return nullable strings

### DIFF
--- a/src/Exceptions/ErrorException.php
+++ b/src/Exceptions/ErrorException.php
@@ -29,7 +29,7 @@ final class ErrorException extends Exception
     /**
      * Returns the error type.
      */
-    public function getErrorType(): string
+    public function getErrorType(): ?string
     {
         return $this->contents['type'];
     }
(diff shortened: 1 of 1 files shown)


gehrisandro:
Hi @nurdism 

Could you please provide an example, when the code or type is empty?

nurdism:
> getErrorCode

I am unsure about `getErrorCode()` because I couldn't find much documentation on *exactly* what comes back in the response, added it to be safer. But as far as `getErrorType()` `429 - You exceeded your current quota, please check your plan and billing details` does not have a type and will throw a type error when attempting to use `getErrorType()`

gehrisandro:
Oh, I just saw that in another issue we have an error response where the code is null: https://github.com/openai-php/client/pull/89#discussion_r1148437425

So both should be nullable.

Please add some tests as @nunomaduro mentioned above and then the PR should be ready.
If you have troubles doing so, please let me know.

nurdism:
I don't even know where to begin to write a test for this, feel free to close this and submit a different PR

gehrisandro:
Okay, no worries. I will do it later this week.

gehrisandro:
Finally found time to fix this https://github.com/openai-php/client/pull/102

PR #89, opened 2023-03-25
Assert valid response in Chat resource for model

IshtarStar:
Improve the error handling in the Chat resource's assertValidResponse method to provide clearer and more informative error messages when a specified model does not exist. This enhancement helps developers to quickly identify and address issues related to non-existent models in their application.

Background to this PR:
There are examples in the documentation, with the use of the model of gpt-4, which is not yet available for everyone through openAI. Moreover, it would be good to take control if a non-existing model is given as a parameter.

openai-playground

My playground:
https://github.com/IshtarStar/openai-playground

### DIFF
--- a/src/Resources/Chat.php
+++ b/src/Resources/Chat.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace OpenAI\Resources;
 
+use Http\Discovery\Exception\NotFoundException;
 use OpenAI\Responses\Chat\CreateResponse;
 use OpenAI\Responses\Chat\CreateStreamedResponse;
 use OpenAI\Responses\StreamResponse;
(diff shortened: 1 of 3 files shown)


src/Resources/Chat.php
```diff
@@ -49,6 +50,26 @@ public function createStreamed(array $parameters): StreamResponse
 
         $response = $this->transporter->requestStream($payload);
 
-        return new StreamResponse(CreateStreamedResponse::class, $response);
+        $streamResponse = new StreamResponse(CreateStreamedResponse::class, $response);
```
nunomaduro, line 53: Can you add some tests?

IshtarStar: Test added.

src/Resources/Chat.php
```diff
@@ -49,6 +50,26 @@ public function createStreamed(array $parameters): StreamResponse
...
+     * Throws a NotFoundException if the status code is 404.
+     *
+     * @param StreamResponse $streamResponse The StreamResponse to be validated
+     * @param string         $model          The model name associated with the StreamResponse
+     *
+     * @throws NotFoundException If the StreamResponse's status code is 404
+     */
+    private function assertValidResponse(StreamResponse $streamResponse, string $model): void
+    {
+        if (404 === $streamResponse->getStatusCode()) {
```
gehrisandro, line 71: To check only against the status code is maybe a bit unsecure. Probably the API will return a 404 in more different states, but they did not document their usage of a 404: 

I think we should check against the response content as the exact error is reported there:

```php
try {
    $response = ($this->streamHandler)($request);
} catch (ClientExceptionInterface $clientException) {
    dd($clientException->getResponse()->getBody()->getContents()); // output see below
    throw new TransporterException($clientException);
}
```

```json
{
    "error": {
        "message": "The model `gpt-42` does not exist",
        "param": null,
        "code": null
    }
}
```

@nunomaduro Maybe we should improve the error handling in general to use the error message returned from the API if available.

gehrisandro (changes requested):
Hi @IshtarStar 

Thank you for raising the issue.

Just as a little clarification: GPT-4 is already available for some users. You can sign-up here to join their waiting list: https://openai.com/waitlist/gpt-4-api

src/Resources/Chat.php
IshtarStar: Currently the error is not coming from the HttpTransporter. I would handle the change iteratively, and once you have reported the bug to OpenAI we can go back to the drawing board. Then you could use the message from the origin API.

gehrisandro: @IshtarStar Sorry, but I don't get it.

What is the error on OpenAI side? They do return a status code 404 which is clearly not a successful response.

Could you please explain how the OpenAI response should look like in your opinion?

IshtarStar: We can close this PR. See new:
https://github.com/openai-php/client/pull/90

PR #90, opened 2023-03-27
Handle client errors in the requestStream method using ResponseClientException

IshtarStar:
This pull request introduces changes to the requestStream method to handle client errors in the 400 range by using a response exception class ResponseClientException that implements the ClientExceptionInterface. This ensures proper error handling while maintaining compatibility with the required interface.

Changes in this PR include:

1. Addition of the ResponseClientException class that implements ClientExceptionInterface.
2. Modification of the requestStream method in the HttpTransporter class to handle client errors in the 400 range by throwing the ResponseClientException.
3. Two new tests to cover scenarios where the requestStream method receives a 401 and a 404 status code.

With these changes, client errors in the 400 range are properly handled, making the requestStream method more robust and reliable.


gehrisandro:
Thank you for clarifying which client you are using.

It looks like Guzzle does not respect the PSR-18 standard properly as it should not throw an exception in case of a 4xx status code: https://www.php-fig.org/psr/psr-18/#error-handling
But anyway we can not change the way Guzzle works.

In my opinion your approach is not too bad but we should not throw different exceptions depending on the HTTP client used. Therefore I would prefer to throw the existing `TransporterException` in both cases.

What do you think?

IshtarStar:
I have adjusted it so that the TransporterException is thrown, independent of the client.

IshtarStar:
@gehrisandro Yes, let's find a solution that you think is appropriate. I know how many users already rely on this exception that it really becomes a breaching change.

If you find time, feel free to look over it for your own way.

gehrisandro:
I did a deeper research on the topic and I think we should unify the exception handling instead of just "fixing" the stream requests: https://github.com/openai-php/client/pull/113

@IshtarStar This would solve your issue too? Right?

PR #106, opened 2023-04-17
feat: http support

ranpro:


ranpro:
This may sound humorous, but in certain regions, it is necessary to use a proxy in order to access OpenAI. The proxy used may not necessarily be utilizing the HTTPS protocol.

debuggerone:
I don't see any humor in using techniques that work against the rules of openai. There are restrictions on available locations for reasons.

nunomaduro:
It's missing tests.

ranpro:
> It's missing tests.

sorry, i just completed the unit tests, thank you.

PR #133, opened 2023-05-31
Fix Json Errors

### LINKED ISSUE #132: ErrorException: Undefined array key "choices"
I'm currently using v0.5.2.

When using the chat completions, I keep getting the following exception when there is a server error on OpenAI's side.

CleanShot 2023-05-31 at 10 51 45@2x

While debugging I came across this line.

CleanShot 2023-05-31 at 10 56 51@2x

Why does it check the charset? Is that required?

If I remove the charset part, then the following exception is thrown which I would expect.

CleanShot 2023-05-31 at 11 02 52@2x

tomirons:
Fixes #132

### DIFF
--- a/src/Transporters/HttpTransporter.php
+++ b/src/Transporters/HttpTransporter.php
@@ -113,7 +113,7 @@ final class HttpTransporter implements TransporterContract
             return;
         }
 
-        if ($response->getheader('Content-Type')[0] !== 'application/json; charset=utf-8') {
+        if (! in_array($response->getheader('Content-Type')[0], ['application/json', 'application/json; charset=utf-8'])) {
             return;
         }
 


src/Transporters/HttpTransporter.php
```diff
@@ -113,7 +113,7 @@ private function throwIfJsonError(ResponseInterface $response, string|ResponseIn
             return;
         }
 
-        if ($response->getheader('Content-Type')[0] !== 'application/json; charset=utf-8') {
+        if (! in_array($response->getheader('Content-Type')[0], ['application/json', 'application/json; charset=utf-8'])) {
```
tomirons, line 116: A simpler fix would be to do this. But I'm uncertain if there consequences.

```suggestion
        if ($response->getheader('Content-Type')[0] !== 'application/json') {
```

gehrisandro:
closed in favour of #134

PR #134, opened 2023-06-01
fix: exception handling for server errors with different content type header - fixes #132

### LINKED ISSUE #132: ErrorException: Undefined array key "choices"

gehrisandro:
This PR fixes exception handling for errors with a. content type header of `application/json` instead of `application/json; charset=utf-8`

fixes #132

### DIFF
--- a/src/Enums/Transporter/ContentType.php
+++ b/src/Enums/Transporter/ContentType.php
@@ -11,4 +11,5 @@ enum ContentType: string
 {
     case JSON = 'application/json';
     case MULTIPART = 'multipart/form-data';
+    case TEXT_PLAIN = 'text/plain';
 }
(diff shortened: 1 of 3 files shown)


tomirons:
This fixes the issue I was encountering.

bmckay959:
Just adding that this also fixed my issue.

`Undefined array key "choices"`

I was trying to send a prompt that would take too many tokens, but was not getting a helpful response until this.

PR #141, opened 2023-06-13
Fixed broken getting of body as a string

awm1:
as in [this](https://github.com/openai-php/client/pull/33) pull request

### DIFF
--- a/src/Transporters/HttpTransporter.php
+++ b/src/Transporters/HttpTransporter.php
@@ -47,13 +47,13 @@ final class HttpTransporter implements TransporterContract
 
         $response = $this->sendRequest(fn (): \Psr\Http\Message\ResponseInterface => $this->client->sendRequest($request));
 
-        $contents = $response->getBody()->getContents();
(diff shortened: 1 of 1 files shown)


awm1:
> @awm1 Could you give an example when does this problem occurs or provide a regression test?

It occurs when I use \OpenAI::factory()->withHttpClient() + Client from php-http/guzzle7-adapter. It should be the same implementation of ClientInterface like Psr18Client, however, it returns empty getContents() due to its internal behaviour mentioned in that PR. I haven't found any standard this client violates so I suppose this fix may be good solution.

gehrisandro:
@awm1 Do you have a working example of a failing setup? Maybe a repository?

The reason I ask: I could merge this PR as I don't see an issue, but I want to ensure it does not get reverted with the next refactoring as it already happened once. Therefore we need some way to write a regression test, which is only possible when I can reproduce the issue.

gehrisandro:
@awm1 Still waiting for your example. But feel free to reopen this PR if needed.

PR #145, opened 2023-06-14
Update CreateResponseFixture.php

benbjurstrom:
Thanks so much for getting the new function stuff added so quickly! 

Found this minor issue while trying to write some tests using CreateResponse::fake. Without the function_call key in the fixture the content of functionCall in the faked response will always be null.

Tested with:
```php
    $client = new ClientFake([
        CreateResponse::fake([
            'choices' => [
                [
                    'message' => [
                        'function_call' => [
                            'name' => 'get_current_weather',
                            'arguments' => "{\n  \"location\": \"Boston, MA\"\n}",
                        ],
                    ],
                    'finish_reason' => 'function_call'
                ],
            ],
        ]),
    ]);
```


gehrisandro:
Thank you for the PR. 🚀

PR #175, opened 2023-07-29
Main

maina-david:
update Psr18Client in client

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -5,6 +5,7 @@ namespace OpenAI;
 use Closure;
 use Exception;
 use GuzzleHttp\Client as GuzzleClient;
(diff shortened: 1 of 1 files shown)


gehrisandro:
@maina-david Could you please provide some background information why this change is needed?

maina-david:
@gehrisandro HttpDiscovery is not an HTTP client itself but rather a library that helps one discover available PSR-18 client implementations in the environment. It allows the package to be more flexible and not tied to a specific HTTP client implementation like Symfony HttpClient.

gehrisandro:
Thank you for clarification. This sounds like a reasonable decision.

Will check if everything is compatible with the other packages and then going to merge it.

PR #178, opened 2023-08-01
Add missing moderation enums.

iBotPeaches:
In testing we were missing a few enums. Traced it back to this package missing some of the enums from here: https://platform.openai.com/docs/guides/moderation/moderation

Linting is failing unrelated to this PR from the main branch as well, so I'll throw another PR for that to keep split.


gehrisandro:
@iBotPeaches Thanks for your PR.

Guess they added more values after we implemented the feature initially.

PR #179, opened 2023-08-01
Fix Lint failures.

iBotPeaches:
While working #178 - my local suite was failing, as well as main [here](https://github.com/openai-php/client/actions/runs/5593233284/job/15150065665). Personally, outside of my knowledge of why since I assume these tools are versions are locked, but there was no package version change on last merge.

Everything passes locally with `composer test` with this.


gehrisandro:
Thanks 👍

PR #184, opened 2023-08-14
Fix function call responses from Azure throwing errors

karlerss-teamdash:
The responses from Azure for gpt-35-turbo-16k don't seem to have the content attribute (with OpenAI it's set to null).

This PR fixes the _ErrorException: Undefined array key "content"_ error.


ukkok:
Just faced same issue and fixed with exactly same solution. Can confirm that this works.

gehrisandro:
@karlerss-teamdash Many thanks for the PR!

PR #203, opened 2023-09-05
fix: change invalid return type of exception

taras-pinchuk-uinno:
Fix invalid return type of OpenAI\\Exceptions\\ErrorException::getErrorCode()

Exception:
TypeError
OpenAI\\Exceptions\\ErrorException::getErrorCode(): Return value must be of type ?string, int returned


gehrisandro:
Hi @taras-pinchuk-uinno 

Thank you. Fixed in main.

PR #204, opened 2023-09-06
fix: Use method that seeks stream

AlfredSorrow:
Right now if you are using a guzzle middleware that reads content (or something else that also reads content), you will get empty string when you call getBody()->getContents() and as a consequence a json decode error, because the stream has already been read

Sample code that reads the body before returning the response:
```php  
$stack = HandlerStack::create();

$stack->push(
    Middleware::log(
        $this->logger,
        new MessageFormatter(MessageFormatter::DEBUG)
    )
);

$this->client = (new Factory())->withApiKey($this->apiKey)
    ->withOrganization($this->organization)
    ->withHttpClient(new Client(['handler' => $stack]))
    ->make();
```  

Solution:  
Use StreamInterface::_toString(), because it must attempt to seek to the beginning of the stream

### DIFF
--- a/src/Transporters/HttpTransporter.php
+++ b/src/Transporters/HttpTransporter.php
@@ -48,7 +48,7 @@ final class HttpTransporter implements TransporterContract
 
         $response = $this->sendRequest(fn (): \Psr\Http\Message\ResponseInterface => $this->client->sendRequest($request));
 
-        $contents = $response->getBody()->getContents();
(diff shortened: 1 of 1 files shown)


gehrisandro:
Hi @AlfredSorrow 

Could you please add a test?

gehrisandro:
Going to close this for now. Feel free to reopen and provide the additional information required.

PR #224, opened 2023-10-08
feat: define default model

justinkekeocha:
Use the specified default model if specified, instead of defining the model on every request


nunomaduro:
Maybe is a good idea, but definitely not via environment variables.

PR #241, opened 2023-11-07
Add support for chat vision response format

gehrisandro:
Add support for GTP-4 vision on the chat completion endpoint: https://platform.openai.com/docs/guides/vision


nunomaduro (approved):
lgtm

PR #253, opened 2023-11-12
Update MetaInformation.php

vesper8:
Handle when x-request-id is not returned, such as when using a local LLAMA model via LM Studio's local server feature.

### DIFF
--- a/src/Responses/Meta/MetaInformation.php
+++ b/src/Responses/Meta/MetaInformation.php
@@ -28,7 +28,7 @@ final class MetaInformation implements MetaInformationContract
      */
     public static function from(array $headers): self
     {
-        $requestId = $headers['x-request-id'][0];
+        $requestId = isset($headers['x-request-id']) ? $headers['x-request-id'][0] : 'n/a';
 
         $openai = MetaInformationOpenAI::from([
             'model' => $headers['openai-model'][0] ?? null,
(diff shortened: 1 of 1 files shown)


src/Responses/Meta/MetaInformation.php
```diff
@@ -28,7 +28,7 @@ private function __construct(
      */
     public static function from(array $headers): self
     {
-        $requestId = $headers['x-request-id'][0];
+        $requestId = isset($headers['x-request-id']) ? $headers['x-request-id'][0] : 'n/a';
```
yaroslav-qlicks, line 31: $requestId = $headers['x-request-id'][0]) ?? 'n/a';

src/Responses/Meta/MetaInformation.php
```diff
@@ -28,7 +28,7 @@ private function __construct(
      */
     public static function from(array $headers): self
     {
-        $requestId = $headers['x-request-id'][0];
+        $requestId = $headers['x-request-id'][0] ?? 'n/a';
```
gehrisandro, line 31:
```suggestion
        $requestId = $headers['x-request-id'][0] ?? null;
```

Instead of `n/a` we should make this field nullable to be consistent with the rest of the package.

And please a test.

gehrisandro (changes requested):
@vesper8 Thanks for your PR.

gehrisandro:
close in favour of https://github.com/openai-php/client/pull/283

PR #255, opened 2023-11-13
Add Error properties to Fine-Tuning Job Response

trippo:
- Add error to Fine-Tuning Job Response
- Update HyperParameters params on Fine-Tuning Job Response

image

### DIFF
--- a/src/Responses/FineTuning/RetrieveJobResponse.php
+++ b/src/Responses/FineTuning/RetrieveJobResponse.php
@@ -12,12 +12,12 @@ use OpenAI\Responses\Meta\MetaInformation;
 use OpenAI\Testing\Responses\Concerns\Fakeable;
 
 /**
- * @implements ResponseContract<array{id: string, object: string, model: string, created_at: int, finished_at: ?int, fine_tuned_model: ?string, hyperparameters: array{n_epochs: int|string, batch_size: int|string, learning_rate_multiplier: int|string}, organization_id: string, result_files: array<int, string>, status: string, validation_file: ?string, training_file: string, trained_tokens: ?int, error: ?array{code: string, param: string, message: string}}>
+ * @implements ResponseContract<array{id: string, object: string, model: string, created_at: int, finished_at: ?int, fine_tuned_model: ?string, hyperparameters: array{n_epochs: int|string, batch_size: int|string|null, learning_rate_multiplier: float|string|null, prompt_loss_weight: float|string|null}, organization_id: string, result_files: array<int, string>, status: string, validation_file: ?string, training_file: string, trained_tokens: ?int, error: ?array{code: string, param: ?string, message: string}}>
  */
 final class RetrieveJobResponse implements ResponseContract, ResponseHasMetaInformationContract
 {
     /**
-     * @use ArrayAccessible<array{id: string, object: string, model: string, created_at: int, finished_at: ?int, fine_tuned_model: ?string, hyperparameters: array{n_epochs: int|string, batch_size: int|string, learning_rate_multiplier: int|string}, organization_id: string, result_files: array<int, string>, status: string, validation_file: ?string, training_file: string, trained_tokens: ?int, error: ?array{code: string, param: string, message: string}}>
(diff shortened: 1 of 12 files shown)


src/Responses/FineTuning/RetrieveJobResponse.php
```diff
@@ -66,6 +67,7 @@ public static function from(array $attributes, MetaInformation $meta): self
             $attributes['validation_file'],
             $attributes['training_file'],
             $attributes['trained_tokens'],
+            RetrieveJobResponseError::from($attributes['error']),
```
gehrisandro, line 70:
```suggestion
            isset($attributes['error']) ? RetrieveJobResponseError::from($attributes['error']) : null,
```

src/Responses/FineTuning/RetrieveJobResponseError.php
```diff
@@ -0,0 +1,56 @@
...
+final class RetrieveJobResponseError implements ResponseContract
+{
+    /**
+     * @use ArrayAccessible<array{code: string, param: string, message: string}>
+     */
+    use ArrayAccessible;
+
+    private function __construct(
+        public readonly string $code,
+        public readonly string $param,
```
gehrisandro, line 22:
```suggestion
        public readonly ?string $param,
```

According to the docs this is nullable: https://platform.openai.com/docs/api-reference/fine-tuning/object

src/Responses/FineTuning/RetrieveJobResponseHyperparameters.php
```diff
@@ -8,29 +8,33 @@
...
     /**
-     * @use ArrayAccessible<array{n_epochs: int|string}>
+     * @use ArrayAccessible<array{n_epochs: int|string, batch_size: int|string, learning_rate_multiplier: int|string}>
      */
     use ArrayAccessible;
 
     private function __construct(
         public readonly int|string $nEpochs,
+        public readonly int|string $batchSize,
+        public readonly int|string $learningRateMultiplier,
```
gehrisandro, line 23: According to the docs this should be a float: https://platform.openai.com/docs/api-reference/fine-tuning/create
```suggestion
        public readonly float|string $learningRateMultiplier,
```

gehrisandro (changes requested):
@trippo Thanks for your PR.

src/Responses/FineTuning/RetrieveJobResponse.php
trippo: i think  `RetrieveJobResponseError::from($attributes['error'] ?? null)` is better

src/Responses/FineTuning/RetrieveJobResponseError.php
trippo: thanks... updated

src/Responses/FineTuning/RetrieveJobResponseHyperparameters.php
trippo: ok on doc `learning_rate_multiplier` can be also null so setted `float|string|null`
I added also `prompt_loss_weight` with same types

src/Responses/FineTuning/RetrieveJobResponse.php
```diff
@@ -66,6 +67,7 @@ public static function from(array $attributes, MetaInformation $meta): self
             $attributes['validation_file'],
             $attributes['training_file'],
             $attributes['trained_tokens'],
+            RetrieveJobResponseError::from($attributes['error'] ?? null),
```
gehrisandro, line 70: Please use isset() to be consistent with the rest of the library!

src/Responses/FineTuning/RetrieveJobResponseHyperparameters.php
```diff
@@ -8,29 +8,35 @@
...
-     * @use ArrayAccessible<array{n_epochs: int|string}>
+     * @use ArrayAccessible<array{n_epochs: int|string, batch_size: int|string|null, learning_rate_multiplier: float|string|null, prompt_loss_weight: float|string|null}>
      */
     use ArrayAccessible;
 
     private function __construct(
         public readonly int|string $nEpochs,
+        public readonly int|string|null $batchSize,
+        public readonly float|string|null $learningRateMultiplier,
+        public readonly float|string|null $promptLossWeight,
```
gehrisandro, line 24:
```suggestion

```

This parameter is only available on the deprecated `FineTune` endpoint and not on `FineTuning`.

gehrisandro (changes requested):

gehrisandro:
@trippo Thank you

PR #283, opened 2023-11-30
fix: handle `x-request-id` in `MetaInformation`

olegnowikow:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Fix handle of the `x-request-id` header when using models that do not return it:
```php
<?php

namespace App\Service;

use OpenAI;

final class Ai
{
    private $client;

    public function __construct()
    {
        $this->client = OpenAI::factory()
            ->withBaseUri('http://localhost:8888/v1')
            ->make();
    }

    public function getModels()
    {
        return $this->client->models()->list();
    }
}

```
image

### Related:

#253

### DIFF
--- a/src/Responses/Meta/MetaInformation.php
+++ b/src/Responses/Meta/MetaInformation.php
@@ -16,7 +16,7 @@ final class MetaInformation implements MetaInformationContract
     use ArrayAccessible;
 
     private function __construct(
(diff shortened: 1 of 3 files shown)


gehrisandro:
Thans @olegnowikow for your PR!

PR #292, opened 2023-12-14
fix: Make object and created in StreamedResponse optional

gerdemann:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

If the endpoint does not return an `object` or `created`, an error no longer occurs.

### DIFF
--- a/src/Responses/Chat/CreateStreamedResponse.php
+++ b/src/Responses/Chat/CreateStreamedResponse.php
@@ -45,8 +45,8 @@ final class CreateStreamedResponse implements ResponseContract
 
         return new self(
             $attributes['id'],
-            $attributes['object'],
-            $attributes['created'],
+            $attributes['object'] ?? '',
+            $attributes['created'] ?? time(),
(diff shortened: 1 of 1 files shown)


gehrisandro:
Hi @gerdemann 

Thank you for your PR. According to the docs, this fields should be always part of the response and are not optional: https://github.com/openai/openai-openapi/blob/master/openapi.yaml#L5632

But sometimes the docs of OpenAI are wrong. Do you have an example how to reproduce the issue?

gerdemann:
Hi @gehrisandro, I tried to connect the Mistral AI models and these fields are missing in the first stream response.
But the documentation doesn't say that either:
https://docs.mistral.ai/api

gehrisandro:
Hi @gerdemann 

Thanks for your quick reply.

As this is a client for the OpenAI API, we are not going to merge this change. But maybe you can reach out to Mistral AI to fix their response, because it's obviously wrong if they try to mirror the OpenAI API.

PR #297, opened 2023-12-19
Switch to supported size for dall-e-3

JeffreyWay:
### Description:

Just a quick readme update. It looks like 256x256 is an invalid size for the dall-e-3 model.

### Related:

See here for supported sizes: https://platform.openai.com/docs/api-reference/images/create#images-create-size


gehrisandro:
Thank you @JeffreyWay 🚀

PR #299, opened 2023-12-20
Update links

Laureano:
Hi! I just noticed that the URLs for the transcription and translation API were outdated.

Thank you for your work on this project!


gehrisandro:
Thank you @Laureano 👍

PR #306, opened 2023-12-22
headers to lowercase

chengxs1994:
### What:

- [x] Bug Fix
- [ ] New Feature


huangdijia (approved):

gehrisandro:
Hi @chengxs1994 

Thank you for your PR!

Could you provide an example why this changes is necessary? As far as I understand the OpenAI docs, the headers should always be in lowercase.

huangdijia:
According to the psr specification, the getHeaderLine method is case insensitive, but the conversion to array does not automatically convert to lowercase

gehrisandro:
Thanks for clarification 👍

Issue #300, opened 2023-12-20
[Bug]: system_fingerprint is always null when using CreateResponse.fake


elliotmassen:
### Description

When I'm creating a fake chat response via `CreateResponse.fake` and I attempt to override the system fingerprint field, it is always set to `null` in the client's response. This means that I'm unable to use the `fake` method to test code that relies on `system_fingerprint`. The current workaround is to use `CreateResponse.from` and provide the full attributes array (including the system fingerprint).

### Steps To Reproduce

```php
<?php

use OpenAI\Responses\Chat\CreateResponse;
use OpenAI\Testing\ClientFake;

// Create a fake client and override the `system_fingerprint` field
$client = new ClientFake([
  CreateResponse::fake([
    'system_fingerprint' => 'abcxyz123'
  ]),
]);

// Make a call to the fake client
$response = $client->chat()->create([
  'model' => 'gpt-4',
  'messages' => [
    ['role' => 'system', 'content' => 'Some system prompt'],
    ['role' => 'user', 'content' => 'Some user prompt']
  ]
]);

// The response's system fingerprint property is `null`
var_dump($response->systemFingerprint);
```

### OpenAI PHP Client Version

v0.8.0

### PHP Version

8.1.24

### Notes

It appears that because the `system_fingerprint` field isn't specified in the `CreateResponseFixture`, it can't be overwritten by `Fakeable.buildAttributes`. It seems that support for `system_fingerprint` was [added in November](https://github.com/openai-php/client/commit/98054e6582d66ebe7f19afa999788f2caf0e16c4#diff-324519387d09305398cfa9dd1fda0641b2e1f66dcf89ceba38c7e3f547d78ca9R75), but it wasn't included in the fixture.

gehrisandro:
Fixed in https://github.com/openai-php/client/pull/308

PR #310, opened 2023-12-22
Add default system_fingerprint to completion create response fixture

elliotmassen:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

This PR continues from #308 by also adding the [`system_fingerprint`](https://platform.openai.com/docs/api-reference/completions/object#completions/object-system_fingerprint) field to the Completions fixture and `CreateResponse` class. This allows the field to be set via the `CreateResponse.fake` method, which has been included in the `"fake with override"` Completion's test.

### Related:

- https://github.com/openai-php/client/issues/300
- https://github.com/openai-php/client/pull/308

### DIFF
--- a/src/Responses/Completions/CreateResponse.php
+++ b/src/Responses/Completions/CreateResponse.php
@@ -32,6 +32,7 @@ final class CreateResponse implements ResponseContract, ResponseHasMetaInformati
         public readonly string $object,
         public readonly int $created,
         public readonly string $model,
+        public readonly ?string $systemFingerprint,
         public readonly array $choices,
         public readonly CreateResponseUsage $usage,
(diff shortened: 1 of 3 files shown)


gehrisandro:
Hi @elliotmassen 

Thank you for your PR, but it looks like this field is not included in a completion response.

I know, the OpenAI docs says, that it should be included, but I was not able to see this field in a test request and I don't think OpenAI has actually added it, as the completion API is already deprecated.

Have you an evidence, that there actually is a `system_fingerprint` field in the response?

gehrisandro:
Closed due to inactivity. Feel free to open a new PR.

PR #319, opened 2024-01-12
Add test case for function call with output not submitted yet

michgeek:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

When a retrieving the list of run step where their results have not submitted yet, it will trigger an error because `output` is missing from the `$attribute` array.


gehrisandro:
@michgeek Thank you for the PR.

Issue #323, opened 2024-01-24
[Bug]: OpenAI\Responses\Files\RetrieveResponse::__construct(): Argument #3 ($bytes) must be of type int, null given


Ruud68:
### Description

Hi,
new user of this wonderful library: many thanks to everybody involved!

I'm running into an issue when using the files endpoint giving me the following error:
```
OpenAI\Responses\Files\RetrieveResponse::__construct(): Argument #3 ($bytes) must be of type int, null given, called in /[xxxxxxxx]/vendor/openai-php/client/src/Responses/Files/RetrieveResponse.php on line 50
```

pinging issue #313 here as that actually isn't solved but working around the issue by not using the files endpoint.

### Steps To Reproduce

1. try to get a list of files:
```
$response = $client->files()->list();
```
2. try to get the file object for a file with id file-xyz
```
$response = $client->files()->retrieve('file-xyz');
```
The organization has files in it and the file with I'm trying to receive exists as I can download it just fine when doing:
```
$client->files()->download(('file-xyz')
```

both 1 and 2 will result in the error.

If anymore information is needed, just ask.

### OpenAI PHP Client Version

v0.8.1

### PHP Version

8.2.15

### Notes
When doing a curl lookup curl https://api.openai.com/v1/files  -H "Authorization: Bearer [my super secret key]"

I get a correct response but do notice that there are files with bytes = null (see snippet below)

```
{
  "object": "list",
  "data": [
    {
      "object": "file",
      "id": "file-X1xx8ghxldIivuxh8uo7kJgm",
      "purpose": "assistants_output",
      "filename": "b5883fa4-24cd-4880-8300-986c9869fb41",
      "bytes": null,
      "created_at": 1706107508,
      "status": "processed",
      "status_details": null
    },
```

eng-ahmad-sameer:
Same issue +1
Any updates about it?
@nunomaduro

Ruud68:
hi @eng-ahmad-sameer 
I have created a PR that will solve this bug and (hopefully) not break anything else.
Can you test it?

gehrisandro:
Fixed in https://github.com/openai-php/client/pull/325

PR #325, opened 2024-01-25
OpenAI\Responses\Files\RetrieveResponse::__construct(): Argument #3 ($bytes) must be of type int, null given

Ruud68:
### What:

- [X] Bug Fix
- [ ] New Feature

### Description:

in the response from OpenAI, bytes can be null. The construct here only accepts integers leading to an error. See bug report for more details (#323)

### Related:
solves reported bug #323


gehrisandro:
@Ruud68 

Thank you for the PR.

PR #343, opened 2024-02-27
Fix Content retrieval in HttpTransport with custom HttpClient

martinhoch42:
`getContent` does not return the content of the body but the remaining content of the stream.

If you were to use `OpenAI::factory` with `withHttpClient` and, for example, a logging middleware, `getContent` would return an empty string since the stream was already read by the log middleware.

See:

https://www.php-fig.org/psr/psr-7/
https://stackoverflow.com/questions/30549226/guzzlehttp-how-get-the-body-of-a-response-from-guzzle-6

fixes: #432


iBotPeaches:
thanks!

PR #345, opened 2024-03-01
Update RetriveJobResponseError

PietroGhislieri98:
### What:

- [ ] Bug Fix
- [ ] New Feature

### Description:

Error Response Api when retriving/creating a training job


iBotPeaches:
Unfortunately due to the change of namespace in `composer.json` and this sitting unreviewed for a bit it is conflicting. If you re-submit please let me know which model you are using and service.

PR #359, opened 2024-03-20
fix json Exception

chengxs1994:
### What:

- [ ] Bug Fix
- [ ] New Feature


chengxs1994:
Throw contents to facilitate locating the cause of failure.

PR #361, opened 2024-03-25
Set default index of 0 if not defined by API server (e.g. OpenRouter)

giltotherescue:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

By setting a different baseUri, this package is fully compatible with OpenRouter.ai, a service that supports many different AI models through a common interface (OpenAI's).

However, OpenRouter does not provide an "index" property in its responses, which causes the response to fail. Setting a default index of 0 fixes this problem.

### DIFF
--- a/src/Resources/Images.php
+++ b/src/Resources/Images.php
@@ -26,7 +26,7 @@ final class Images implements ImagesContract
     {
         $payload = Payload::create('images/generations', $parameters);
 
-        /** @var Response<array{created: int, data: array<int, array{url?: string, b64_json?: string, revised_prompt?: string}>}> $response */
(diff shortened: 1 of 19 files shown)


cloudinstone:
why this request is not merged? I try to use this client with my openrouter api.  it's caused the error:
Warning: Undefined array key "choices" in /Volumes/G-Drive/Local Sites/Test/app/public/wp-content/plugins/wp-cake/vendor/openai-php/client/src/Responses/Chat/CreateResponse.php on line 50

iBotPeaches:
Sorry for delay. Marking as draft due to conflicts.

iBotPeaches:
Something went odd with the rebase/merge here.

giltotherescue:
I believe this has been addressed in other ways in newer versions, so I am closing this PR. Thanks

Issue #362, opened 2024-03-29
[Bug]: Warning: Undefined array key "code" in Responses/FineTuning/RetrieveJobResponseError.php on line 42


alek-s-andr:
### Description

A am getting the error when run any method on the `$openAIClient->fineTuning()` endpoint, for example `$openAIClient->fineTuning()->createJob($params);`.
This run leads to fatal error:
`PHP Fatal error:  Uncaught TypeError: OpenAI\Responses\FineTuning\RetrieveJobResponseError::__construct(): Argument #1 ($code) must be of type string, null given, called in vendor/openai-php/client/src/Responses/FineTuning/RetrieveJobResponseError.php on line 37 and defined in vendor/openai-php/client/src/Responses/FineTuning/RetrieveJobResponseError.php:20`

![2024-03-29-13-14-15](https://github.com/openai-php/client/assets/34573954/66640c81-fc5e-4a48-b838-a5e62b9c4d75)

### Steps To Reproduce

Run `$openAIClient->fineTuning()->retrieveJob('ftjob-...');`

### OpenAI PHP Client Version

v0.8.4

### PHP Version

8.1.27

### Notes

https://platform.openai.com/docs/api-reference/fine-tuning/object
I did a little research and found out that the returned **fine-tuning job object** from OpenAI API actually contains the parameter **error** which contains other object with key **error** (but this is not mentioned in the OpenAI documentation) in the absence of an error:

![2024-03-29-12-44-08](https://github.com/openai-php/client/assets/34573954/39ea3e5f-72eb-447c-804a-4e12b1587397)

alek-s-andr:
I created a PR with a fix for this issue: https://github.com/openai-php/client/pull/363

BrainPackIO:
Here is how you can create a completely new install that recreates this error. I have also included the fix

php -v:

PHP 8.2.12 (cli) (built: Oct 24 2023 21:15:15) (ZTS Visual C++ 2019 x64)
Copyright (c) The PHP Group
Zend Engine v4.2.12, Copyright (c) Zend Technologies

# composer require openai-php/client

# composer dump-autoload    

# nano index.php

Paste this:

<?php
require __DIR__ . '/vendor/autoload.php';

$client = OpenAI::client('insert-your-key-here');

$response = $client->fineTuning()->listJobs();
?>

# php index.php

Error Response:

[...]

PR #363, opened 2024-03-29
#362 Fix for the error key in RetrieveJobResponseError in Fine-Tuning

alek-s-andr:
### What:

- [ ] Bug Fix

### Description:

Fix for the issue #362


BrainPackIO:
Here is how you can create a completely new install that recreates this error. I have also included the fix:

https://github.com/openai-php/client/issues/362#issuecomment-2032955368

iBotPeaches (changes requested):
Sorry for delay. I ran the CI on this change and its failing. I can take a look again when its green/passing.

iBotPeaches:
This was fixed in https://github.com/openai-php/client/pull/499 - closing.

PR #367, opened 2024-04-02
Implement Assistants Streaming

EthanBarlo:
### What:

- [ ] Bug Fix
- [X] New Feature

### Description:

This PR implements Full Assistants Streaming API support.
https://platform.openai.com/docs/api-reference/assistants-streaming/events

- ThreadMessageDelta (Including nested classes)
- ThreadRunStepDelta (Including nested classes)

- StreamedThreadRunResponseFactory (To map the events to the correct classes)
- EventStreamResponse (Modified iterator to work with the events format)
- EventStreamResponseItem (Contains the event type and the data, similar to how the api returns the data)

### Related:

https://github.com/openai-php/client/issues/357

PR #392, opened 2024-05-16
Handle missing attributes for assistant streaming

robbie-thompson:
### What:

- [X] Bug Fix
- [ ] New Feature

### Description:

Fixes an error when OpenAI's API returns a delta object which has empty tool calls:

`Undefined array key "id" {"exception":"[object] (ErrorException(code: 0): Undefined array key \"id\" at /var/www/html/vendor/openai-php/client/src/Responses/Threads/Runs/Steps/ThreadRunStepResponseFunctionToolCall.php:41)`

```
(
    [id] => step_voklimBxRmx023PqaIxX6hya
    [object] => thread.run.step.delta
    [delta] => Array
        (
            [step_details] => Array
                (
                    [type] => tool_calls
                    [tool_calls] => Array
                        (
                            [0] => Array
                                (
                                    [index] => 0
                                    [type] => function
                                    [function] => Array
                                        (
                                            [arguments] => {}
                                        )
                                )
                        )
                )
        )
)
```

This is happening to me for all streamed assistant responses that invoke function calls, the function itself appears in one delta, and the next delta then shows the empty function above.

The current implementation is expecting these delta objects to have ID and name fields, which it seems the empty functions don't have.

This is related to the assistants streaming functionality merged in earlier today: https://github.com/openai-php/client/pull/367/

### DIFF
--- a/src/Responses/Threads/Runs/Steps/Delta/ThreadRunStepDeltaResponse.php
+++ b/src/Responses/Threads/Runs/Steps/Delta/ThreadRunStepDeltaResponse.php
@@ -21,7 +21,7 @@ final class ThreadRunStepDeltaResponse implements ResponseContract
     use Fakeable;
 
     private function __construct(
-        public string $id,
+        public ?string $id,
         public string $object,
         public ThreadRunStepDeltaObject $delta,
     ) {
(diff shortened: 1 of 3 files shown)


KaipiYann:
Thks @robbie-thompson  !
I had the same problem...And your fix is working !

gehrisandro:
Thanks @robbie-thompson for the PR!

PR #396, opened 2024-05-19
Fixing CreateResponseChoice text option for GPT4

Dlas1212:
Updating CreateResponseChoice to support gpt-4o model's responses

### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Using the new GPT 4o model, you can't use the `completion()->createStreamed()` method since response is now different. 

A) The base API Url needs to be manually set to `https://api.openai.com/v1/chat` when declaring a new client. 
B) The response from the completion includes "choices", but using GPT-4o model the text result now comes from `$attributes['delta']['content']` instead of `$attributes['test']`

### DIFF
--- a/src/Responses/Completions/CreateResponseChoice.php
+++ b/src/Responses/Completions/CreateResponseChoice.php
@@ -15,15 +15,15 @@ final class CreateResponseChoice
     }
 
     /**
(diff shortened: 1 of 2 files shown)


gehrisandro:
Hi @Dlas1212 

The completions() endpoint is outdated and you have to use the chat() endpoint now. See: https://platform.openai.com/docs/api-reference/completions

PR #413, opened 2024-05-27
Remove inaccurate comment

sebapastore:
### Description:

Just remove an inaccurate comment I found while digging into the source code.


gehrisandro:
Thx @sebapastore

PR #416, opened 2024-06-04
Added filename support for the file payload

apc-harold:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

When uploading a file using:
```
$client->files()->upload([
    'purpose' => $purpose,
    'file' => $file,
]);
```
The uploaded file is automatically named `Nyholm-Psr7-Zval:`, which causes issues because it doesn't have a file extension. This prevents the file from being attached to an assistant.

The proposed changes allow for a file name to be provided by instead passing an array value for the file key:
```
$client->files()->upload([
    'purpose' => $purpose,
    'file' => [
        'data' => $file,
        'fileName' => $fileName,
    ],
]);
```

### DIFF
--- a/src/ValueObjects/Transporter/Payload.php
+++ b/src/ValueObjects/Transporter/Payload.php
@@ -140,6 +140,8 @@ final class Payload
 
     /**
      * Creates a new Psr 7 Request instance.
+     *
+     * @throws ErrorException
      */
     public function toRequest(BaseUri $baseUri, Headers $headers, QueryParams $queryParams): RequestInterface
     {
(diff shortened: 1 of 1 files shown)


gehrisandro:
Hi @apc-harold 

I am not able to reproduce the issue. When I upload a file, the filename is set properly.

Could you please provided a full example or a demo repository which reproduces the issue?

apc-harold:
Hi @gehrisandro,

Thanks for checking it out!

I'm quite busy at the moment so won't be able to provide an example, but:
- Using Laravel, using $file->get() to retrieve the file
- Using the [beta version](https://github.com/openai-php/client/releases/tag/v0.10.0-beta.1)

I'll maybe have some time on w/s 17th, so will make a reminder to come back and make an example if it's still here.

DennisBirkholz:
I stumbled on the same problem:
it is not easily possible to upload string data as a file, because no filename (or an invalid file name) gets sent that is missing the ".jsonl" extension.

This may also be remedied by adding the following code to the file upload example:
```PHP
$jsonlData = "{...}\n{...}";
$response = $client->files()->upload([
        'purpose' => 'fine-tune',
        'file' => \GuzzleHttp\Psr7\Utils::streamFor($jsonlData, ["metadata" => ["uri" => "upload.jsonl"]]),
    ]);
```

olimarferraz:
@DennisBirkholz thanks! Works for me. I'm using the Laravel driver and was stuck in how properly handle this uploads.

PR #422, opened 2024-06-07
fix: Update image url content type to use url instead of file id

knash94:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

This fixes an issue where the image url content type response returns a `url` field instead of `file_id` which was previously put into the code. Updated the existing test coverage to use the correct response.

[Open AI docs for message object](https://platform.openai.com/docs/api-reference/messages/object)


gehrisandro:
Thanks a lot @knash94

PR #427, opened 2024-06-15
Update gpt-4 to gpt-4o on readme

trippo:


iBotPeaches:
Sorry for delay on this. I almost think `gpt-4o-mini` is now the play for a good default in readme, what do you think?

iBotPeaches:
thanks! merging anyway, 4o is cheaper and a better default for readme

PR #429, opened 2024-06-17
Introduce OpenAIThrowable

VincentLanglet:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Hi @gehrisandro 

I'm trying to move https://github.com/openai-php/client/pull/287 forward by splitting it in 3 smaller PRs:
- First PR to introduce OpenAIThrowable
- Second PR to introduce UnreadableResponse exception
- Third PR with only Phpdoc update

The purpose of this PR is to provide an easy way to try/catch method of this lib:
- It avoids catching `\Exception` which catch everything
- It avoids catching `OpenAI\Exceptions\ErrorException|OpenAI\Exceptions\InvalidArgumentException|...` which is complicated and can miss an exception if a new one is introduced later.

### Related:

https://github.com/openai-php/client/pull/287

### DIFF
--- a/src/Exceptions/ErrorException.php
+++ b/src/Exceptions/ErrorException.php
@@ -6,7 +6,7 @@ namespace OpenAI\Exceptions;
 
 use Exception;
 
-final class ErrorException extends Exception
+final class ErrorException extends Exception implements OpenAIThrowable
 {
     /**
      * Creates a new Exception instance.
--- a/src/Exceptions/InvalidArgumentException.php
+++ b/src/Exceptions/InvalidArgumentException.php
@@ -6,4 +6,4 @@ namespace OpenAI\Exceptions;
 
 use Exception;
 
-final class InvalidArgumentException extends Exception {}
+final class InvalidArgumentException extends Exception implements OpenAIThrowable {}
--- a/src/Exceptions/OpenAIThrowable.php
+++ b/src/Exceptions/OpenAIThrowable.php
@@ -1,0 +1,9 @@
+<?php
+
+namespace OpenAI\Exceptions;
+
+use Throwable;
+
+interface OpenAIThrowable extends Throwable
+{
+}
--- a/src/Exceptions/OpenAIThrowable.php
+++ b/src/Exceptions/OpenAIThrowable.php
@@ -4,6 +4,4 @@ namespace OpenAI\Exceptions;
 
 use Throwable;
 
-interface OpenAIThrowable extends Throwable
-{
-}
+interface OpenAIThrowable extends Throwable {}
--- a/src/Exceptions/TransporterException.php
+++ b/src/Exceptions/TransporterException.php
@@ -7,7 +7,7 @@ namespace OpenAI\Exceptions;
 use Exception;
 use Psr\Http\Client\ClientExceptionInterface;
 
-final class TransporterException extends Exception
+final class TransporterException extends Exception implements OpenAIThrowable
 {
     /**
      * Creates a new Exception instance.
--- a/src/Exceptions/UnknownEventException.php
+++ b/src/Exceptions/UnknownEventException.php
@@ -6,4 +6,4 @@ namespace OpenAI\Exceptions;
 
 use Exception;
 
-final class UnknownEventException extends Exception {}
+final class UnknownEventException extends Exception implements OpenAIThrowable {}
--- a/src/Exceptions/UnserializableResponse.php
+++ b/src/Exceptions/UnserializableResponse.php
@@ -7,7 +7,7 @@ namespace OpenAI\Exceptions;
 use Exception;
 use JsonException;
 
-final class UnserializableResponse extends Exception
+final class UnserializableResponse extends Exception implements OpenAIThrowable
 {
     /**
      * Creates a new Exception instance.


VincentLanglet:
Friendly ping @nunomaduro @gehrisandro, do you have some time for a review ? Thanks.

VincentLanglet:
Friendly ping @nunomaduro @gehrisandro,

Is something missing for a review ?
Since https://github.com/openai-php/client/pull/287 got interest but no reaction,
I'm trying to moving it forward with smaller PR. Please tell me if I can do it better :) 

Thanks

VincentLanglet:
Anyway to move forward this PR @gehrisandro @nunomaduro ?

Thanks a lot

VincentLanglet:
Anyway to move forward this PR @gehrisandro @nunomaduro ?

Thanks a lot

iBotPeaches (commented):
Sorry for delay. I took a look at this and outside of the linting failing - I'm not really following this implementation.

* It appears to introduce a custom interface (OpenAIThrowable) that extends Throwable
* All classes then implement that interface.

Could you give any background to help me learn why you took that route instead of say like a base class of like OpenAiException that all these others extended from, which then extended Exception.

It seems like on a face value there is just a bit of workarounds here to avoid this callout. 

> PHP classes cannot implement the Throwable interface directly, and must instead extend [Exception](https://www.php.net/manual/en/class.exception.php).

I want to be sure I'm understanding any trade-offs properly.

VincentLanglet:
Hi @iBotPeaches, thanks for your review.

My main goal is 
- to update the phpdoc from the method of this repository to add `@throws`
- Regroup all the exceptions with a base class/interface in order to be able to `catch (BaseClassOrInterface $e) {}`

I mostly saw in such situation the following strategy:
- Introducing an interface (or many) like https://github.com/symfony/symfony/tree/7.3/src/Symfony/Contracts/HttpClient/Exception
- then, implementing this interface https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/HttpClient/Exception/ClientException.php

Introducing an Interface offers more flexibility because I can write
```
- MyLogicException extends \LogicException implements CustomThrowable
- MyRuntimeException extends \RuntimeException implements CustomThrowable
```
For instance, and would allow to
- Just catch CustomThrowable
- Just catch RuntimeException instead

Notice I can also implements multiple interfaces when I cannot extends multiple classes. So it offer more flexibility.

But I know some repository prefer to introduce a BaseException (or used this strategy because it was introduced before the Throwable interface exists in PHP). If it's the preferred choice from OpenAI, I won't go against it if it the preferred choice.

https://github.com/openai-php/client/pull/287 didn't get any review in 2 years and this PR is waiting for 1 year so I'd like to move on this in order to "fix" the phpdoc and adding the `@throws` annotation :)

VincentLanglet:
@iBotPeaches I rebased so you can approval the workflow run.

Also I'll wait for your decision if you prefer a baseException or an Interface

iBotPeaches:
I kicked off the workflows and I don't have a strong opinion on base vs interface. Let me do some research though to double check if I'm missing anything.

iBotPeaches:
I took a look at your game plan and have a few thoughts I'm looking for clarity on.

* I grepped codebase and I see `@throws` 3 times, attached right to the agnostic HTTP client implementation for all the ways this can fail. The classes after this change would have the interface and for folks that want to just to catch 1 thing - they can still do that. So maybe you add the interface for documentation into the tag, but that seems weird right? You'd have the base interface and then the concrete implementations which extends that base, which still feels weird. You wouldn't add `throw` tags with the unified interface to all methods, because it would be invalid. Like the stream methods can only throw a subset of the errors that the non-streaming can.

* If you are just changing your code to capture any and all OpenAI Exception (this new thing) - you are probably the same type of engineer to just capture any exception during the execution of this library. I'm sure a 

* UnreadableResponse exception - what is use-case for this? I'd argue the issues we have right now is the code is actually crashing when its not json. I'd argue that its "UnserializableException", but we can't even throw it because we crash out (null, string, etc) before getting there.

I wanted to give this PR the time it deserved after waiting a solid few years, but with the bugs and missing parity with new OpenAI APIs - unfortunately at the moment re-hauling the exceptions/throws which can have a direct impact on strict tooling for user-land code does not seem like something I should focus at the moment.

VincentLanglet:
> * UnreadableResponse exception - what is use-case for this? I'd argue the issues we have right now is the code is actually crashing when its not json. I'd argue that its "UnserializableException", but we can't even throw it because we crash out (null, string, etc) before getting there.

It was made to catch
```
$response->getBody()->getContents();
```
https://github.com/openai-php/client/pull/287/files#diff-0344a87ebdb10e227d2ba6b93d934fe88e0641c95e5d443f90436e3c8972cab9R54-R58

If we introduce an interface or a baseException, to allows people catching it rather than
```
catch (ExceptionA|ExceptionB|ExceptionC|...) {}
```
people will still need to be aware of the implementation of OpenAIThrowable, cause it might throw a `RuntimeException`.

So instead it's recommended to throw a Domain-related Exception.

> If you are just changing your code to capture any and all OpenAI Exception (this new thing) - you are probably the same type of engineer to just capture any exception during the execution of this library. I'm sure a

"I'm sure a", is it missing the end of the sentence @iBotPeaches ?

> I grepped codebase and I see @throws 3 times, attached right to the agnostic HTTP client implementation for all the ways this can fail. The classes after this change would have the interface and for folks that want to just to catch 1 thing - they can still do that. So maybe you add the interface for documentation into the tag, but that seems weird right? You'd have the base interface and then the concrete implementations which extends that base, which still feels weird. You wouldn't add throw tags with the unified interface to all methods, because it would be invalid. Like the stream methods can only throw a subset of the errors that the non-streaming can.

If you prefer I can add specific `@throws` tag everywhere without introducing the interface. My Main goal is to have a well documented `@throws` tag and it's way easier when you use an interface.

But if you prefer having 
```
@throws ErrorException
@throws InvalidArgumentException
@throws TransporterException
@throws UnserializableResponse
@throws RuntimeException
```
almost everywhere rather than `@throws OpenAiThrowable` I can do the PR.

bytestream:
>  you are probably the same type of engineer to just capture any exception during the execution of this library. 

They're very different things catching `OpenAIThrowable` and `\Throwable` or `\Exception`. The interface is most beneficial when there's a method that's going to throw multiple exceptions defined in this library. It's excessive to define each individually in a catch, when realistically the catch behaviour is most likely the same in all scenarios that this lib throws.

VincentLanglet:
If preferred @iBotPeaches, I can open a PR with just all the `@throws` added (and validated by PHPStan)
https://github.com/openai-php/client/compare/main...VincentLanglet:client:phpdoc?expand=1

bytestream:
Am I missing something? The PR changes 6 files and effects no one from a BC perspective, it literally just adds an interface. 

The primary sticking point seems to be interface vs base class, and I don't think there's a particularly hard opinion either way.

iBotPeaches:
I guess my confidence is the only thing. For all the other PRs thus far I can sit down and replicate a bug / test a feature. Here I am traveling through more of a design pattern than a bug/feature in my understanding.

Thus with a discussion of patterns (throwable/exception) and preference (interface/base exception) both which are a jumping point for further changes I wanted to type it out a bit to try and understand why this PR sat.

Issue #433, opened 2024-06-20
[Bug]: \OpenAI\Resources\VectorStoresFileBatches::cancel should POST, missing `cancel` path component


jhariani:
### Description

If you attempt to use `\OpenAI\Resources\VectorStoresFileBatches::cancel` to cancel a vector store file batch, the OAI API will return this error:
```
Invalid method for URL (DELETE /v1/vector_stores/vs_id/file_batches/vsfb_id)
```
This is because the payload is constructed with this HTTP method and URL pattern:
```
$payload = Payload::delete("vector_stores/$vectorStoreId/file_batches", $fileBatchId);
```
Per the [OAI cancel batch documentation](https://platform.openai.com/docs/api-reference/vector-stores-file-batches/cancelBatch) it should POST to cancel the batch, and construct the URL with this pattern:

```
https://api.openai.com/v1/vector_stores/{vector_store_id}/file_batches/{batch_id}/cancel
```
Solution: Change [this](https://github.com/openai-php/client/blob/e9ca2886aa9c0bf60fb3b8df00ad9e5b876a0916/src/Resources/VectorStoresFileBatches.php#L73) to use the Payload's cancel helper method:
```
$payload = Payload::cancel("vector_stores/$vectorStoreId/file_batches", $fileBatchId);
```

### Steps To Reproduce

1. Create a file```$file = app(\OpenAI\Client::class)->files()->upload($data);```
2. Create a vector store file batch ```$batch = app(\OpenAI\Client::class)->vectorStores()->batches()->create($vectorStoreId, ['file_ids' => [$file->id]);```
3. Cancel it with ```app((\OpenAI\Client::class)->vectorStores()->batches()->cancel($vectorStoreId, $batch->id)```

<img width="1347" alt="Screenshot 2024-06-20 at 1 39 56 PM" src="https://github.com/openai-php/client/assets/688874/ac13c624-0c5e-4b9d-b776-0ee59f03d761">

### OpenAI PHP Client Version

v0.10.1

### PHP Version

8.2.17

PR #435, opened 2024-06-21
fix: vector stores file batch cancel payload

TyperEJ:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:
\OpenAI\Resources\VectorStoresFileBatches::cancel should POST

### Related:
#433 
[OpenAI API Document](https://platform.openai.com/docs/api-reference/vector-stores-file-batches/cancelBatch)


gehrisandro:
Thanks a lot @TyperEJ

Issue #457, opened 2024-07-26
[Bug]: Undefinded array key "created" in src/Responses/Models/RetrieveResponse


BjornValkOptimizers:
### Description

When using the Azure OpenAI, calling $client->models()->list() will result in a "Undefined array key "created" exception thrown by src/Responses/Models/RetrieveResponse.

### Steps To Reproduce

create a client using:
$client = OpenAI::factory()
    ->withBaseUri('{your-resource-name}.openai.azure.com/openai')
    ->withHttpHeader('api-key', '{your-api-key}')
    ->withQueryParam('api-version', '{version}')
    ->make();

fetch and output the response of models list using $client->models()->list()

### OpenAI PHP Client Version

v0.10.1

### PHP Version

8.1.0

the-fermi-paradox:
This should also have been closed by https://github.com/openai-php/client/pull/567

iBotPeaches:
Closed in https://github.com/openai-php/client/pull/567

PR #458, opened 2024-07-26
Fix undefined array key "created"

BjornValkOptimizers:
Fixed the instance of Models RetrieveResponse not containing 'created' attribute when using Azure OpenAI

### What:

- [X ] Bug Fix
- [ ] New Feature

### Description:

Fix an exception when calling client->models()->list() with an Azure OpenAI configuration.
The Azure OpenAI does not return _created_ and _owned_by_ in it's response. Instead use _created_at_ for the created property and an empty string for the ownedBy property

### Related:

https://github.com/openai-php/client/issues/457


xin-admin:
strongly necessary

iBotPeaches (changes requested):
I'm a bit torn on this. If this is a specific difference of OpenAI on Azure vs main, then we should talk about making some of these nullable.

Since keeping the typing happy with just check after check to fallback to a 0 doesn't seem right. Either way CI failed because the `??` aren't going to work when its typed to always be there.

PR #467, opened 2024-08-20
chore: handle all UnhandledMatchErrors

chris-lee-lb:
### What:

- [X] Bug Fix
- [ ] New Feature

### Description:

Our sentry has been logged many times **UnhandledMatchError** with class ThreadRunStepResponseToolCallsStepDetails.

截圖 2024-08-20 下午6 57 08

So this PR try to throw wrapped exception like **ThreadRunStreamResponse** does for all matches.

### Related:

N/A

### DIFF
--- a/src/Exceptions/UnknownTypeException.php
+++ b/src/Exceptions/UnknownTypeException.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenAI\Exceptions;
+
+use Exception;
+
+final class UnknownTypeException extends Exception
+{
+    public function __construct(private string $unknownType)
+    {
+        parent::__construct('Unknown type: '.$this->unknownType);
+    }
+
+    public function getType(): string
+    {
+        return $this->unknownType;
+    }
+}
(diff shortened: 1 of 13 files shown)


gehrisandro:
Hi @chris-lee-lb 

Thank you for your PR.

In my opinion, it would be better to handle the types properly instead of throwing an exception.

The problem is, that according to the specs, there shouldn't be a different type. 

Could you please provide some steps to reproduce the issue?

chris-lee-lb:
@gehrisandro sorry for later reply. 

And unfortunately we can not reproduce the issue easily, we only occasionally see this error in our sentry platform (https://www.sentry.io), but the sentry doesn't record what http response was actually received, so I can't give you any further debug information.

So my intention of this PR is to use the new Exception to allow the sentry to record the unprocessed type content, so that it can be continuously corrected.

iBotPeaches:
Not going to move forward with this. It makes our typing a mess with unknown - the crux of the issue is not having the source payload available to exceptions which is being worked on actively.

Sorry for delay, but paired with Assistants deprecated. I don't see moving on this.

PR #471, opened 2024-08-25
Fix 'attachments' key exception on thread message modify

### LINKED ISSUE #470: [Bug]: Modify thread message throws 'attachments'  key exception
### Description

When modifying a thread message's metadata after creation, the call to do so fails with:

```
  ErrorException

  Undefined array key "attachments"

  at vendor/openai-php/client/src/Responses/Threads/Messages/ThreadMessageResponse.php:64
     60▕         );
     61▕
     62▕         $attachments = array_map(
     63▕             fn (array $attachment): ThreadMessageResponseAttachment => ThreadMessageResponseAttachment::from($attachment),
  ➜  64▕             $attributes['attachments']
     65▕         );
     66▕
     67▕         return new self(
     68▕             $attributes['id'],
```

### Steps To Reproduce

```
$thread = $client->threads()->create([]);

$message = $client->threads()->messages()->create($thread->id, [
    'role' => 'user',
    'content' => 'Explain deep learning to a 5 year old.',
]);

$response = $client->threads()->messages()->modify(
    threadId: $thread->id,
    messageId: $message->id,
    parameters: [
        'metadata' => [
            'key' => 'value',
        ],
    ]
);
```

### OpenAI PHP Client Version

dev-main

### PHP Version

8.3.9

### Notes

_No response_

datashaman:
### What:

- [X] Bug Fix
- [ ] New Feature

### Description:

Modifying a thread message after creation throws an _attachments_ key exception. This uses the null coalescing operator to provide a fallback if the value is missing.

### Related:

Resolves #470

### DIFF
--- a/src/Responses/Threads/Messages/ThreadMessageResponse.php
+++ b/src/Responses/Threads/Messages/ThreadMessageResponse.php
@@ -61,7 +61,7 @@ final class ThreadMessageResponse implements ResponseContract, ResponseHasMetaIn
 
         $attachments = array_map(
             fn (array $attachment): ThreadMessageResponseAttachment => ThreadMessageResponseAttachment::from($attachment),
-            $attributes['attachments']
+            $attributes['attachments'] ?? []
         );
 
         return new self(


gehrisandro:
Thanks, @datashaman for the PR!

PR #476, opened 2024-09-13
Fixed attachments exception on ThreadMessageResponse.php

misteryomi:
### What:

- [-] Bug Fix

### Description:
The Thread Message returns an exception error when empty. Based on the Open AI docs in https://platform.openai.com/docs/api-reference/messages/object, the `attachment` param must either be an array|null.

The bug fix has added an extra check to allow for the `attachment` to receive a null value (or not throw an exception if not set).

### Related:

Resolves: https://github.com/openai-php/client/issues/470


gehrisandro:
Thanks @misteryomi 

This has already been fixed in a different PR.

Issue #456, opened 2024-07-25
[Bug]: Undefined array key "file_id


ckefalianou:
### Description

 **Undefined array key "file_id"**
 
I did some digging, and found out that the issue is created in that file: **openai-php/client/src/Responses/Threads/Messages/ThreadMessageResponseContentImageUrl.php**

because it looks for **file_id** not **url** , which **file_id** doesn't exist in the content I sent.

Can you please resolve it or suggest a solution?
Thank you in advance!

### Steps To Reproduce

I am using the library, sending a **text** and some **image_url** as content.
After retrieving the Assistant, creating a Thread, I create the following Message:

`        $openAI_message = $client->threads()->messages()->create($openAI_thread->id, [
            'role' => 'user',
            'content' => [
                [
                    'type' => 'text',
                    'text' => $prompt['title']
                ],
                [
                    'type' => 'image_url',
                    'image_url' => [
                        'url' => $photo[0]->large,
                        'detail' => 'high',
                    ],
                ],
                [
                    'type' => 'image_url',
                    'image_url' => [
                        'url' => $photo[1]->large,
                        'detail' => 'high',
                    ],
                ],
            ]
        ]);
`
After creating the Run I get the following issue: Undefined array key "file_id"

### OpenAI PHP Client Version

^0.10.1

### PHP Version

8.3

SergiuDihel:
I hit pretty much the same issue.

I am trying to do something like:

```php
'thread' => [
    'messages' => [[
        'role' => 'user',
        'content' => [
            [ 'type' => 'text', 'text' => $description ],
            [ 'type' => 'image_url', 'image_url' => [ 'url' => 'https://example.com/image.jpg']],
        ],
    ]],
],
```

From what I can understand, the issue is originating inside `src/Responses/Threads/Messages/ThreadMessageResponseContentImageUrlObject.php` on line 41. 

```php
return new self(
    $attributes['type'],
    ThreadMessageResponseContentImageUrl::from($attributes['image_url']),
);
```

`ThreadMessageResponseContentImageUrl::from($attributes['image_url'])` is trying to send a string (probably) containing the URL of the image. But if you look inside the static `from`, in `ThreadMessageResponseContentImageUrl`, you'll see it expects an `array` of attributes:

```php
public static function from(array $attributes): self
{
    return new self(
        $attributes['file_id'],
        $attributes['detail'] ?? null,
    );
}
```

Șo it fails.
[...]

PR #477, opened 2024-09-18
fix critical issue: parsing messages with image_url

dpetrovaliev:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

I've experienced issue while using the #openai-php/laravel repo which depends on this repo.
The issue was related to messages with content type image_url instead of image_file. Parsing this type of file within the current structure of the project is wrong. I assume it's copy/paste issue because the structure of the classes that parse image_file response was the same as these for image_url response. 

With the current PR I'm solving this issue.

Here is the API reference:
https://platform.openai.com/docs/api-reference/messages/createMessage#messages-createmessage-content

### Related:
https://github.com/openai-php/client/issues/456
https://github.com/openai-php/laravel/issues/107
https://github.com/openai-php/laravel/issues/116


gehrisandro:
Already fixed in previous release.

PR #487, opened 2024-10-11
HTTP status in ErrorException

GuidoBelluomo:
### What:
- [ ] Bug Fix
- [X] New Feature

### Description:

I've added the HTTP status code to ErrorException, so it's possible to ignore 404 errors for deleting files that don't exist, if desired. OpenAI doesn't provide an exhaustive enough error response for missing files that I can just quickly and universally check:

```json
{
  "error": {
    "message": "No such File object: file-testFile",
    "param": "id",
    "code": null
  }
}
```

As you can see there's no easy error code for missing resources, parsing the message is a terrible idea, so exposing the HTTP status code (404 in this case) enables to perform such simple checks.


gehrisandro:
Hi @GuidoBelluomo 

Thanks for you PR!

PR #488, opened 2024-10-13
Made some fields in audio transcrtiption response nullable for compatibility with "OpenAI API Compatible" Local-AI API

detain:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Made seek/temperature/avg_logprob/compression_ratio/no_speech_prob fields in audio transcription response nullable for compatibility with using Local-AI API


gehrisandro:
Hi @detain 

We are not going to support APIs other than the official OpenAI API.

Issue #492, opened 2024-11-05
[Bug]: CreateResponseUsage is missing support for Prompt Caching


ReSpawN:
### Description

`\OpenAI\Responses\Chat\CreateResponseUsage` doesn't honor the Prompt Caching return which causes confusion that ChatGPT isn't 'activating' Prompt Caching (perhaps due to a unsupportive model).

```json
{"prompt_tokens":1669,"completion_tokens":526,"total_tokens":2195,"prompt_tokens_details":{"cached_tokens":1152},"completion_tokens_details":{"reasoning_tokens":0,"accepted_prediction_tokens":0,"rejected_prediction_tokens":0}}
```

### Steps To Reproduce

Simply produce a request to chat/completions (`chat()->create([])`).

### OpenAI PHP Client Version

v0.10.2

### PHP Version

8.3.10

gehrisandro:
Implemented in https://github.com/openai-php/client/pull/494

PR #494, opened 2024-11-05
fix: add missing properties (specially `cached_usage`) to chat `CreateResponseUsage`

IsraelOrtuno:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:
Chat `CreateResponseUsage` object is omitting some properties which have probably been have added in more modern versions of the API which are related to the tokens usage. I think `cached_usage` is specially important (two issues were already requesting this) as this would help users track if their prompts are being cached properly and could directly contribute to save money.

Check the `usage` property for reference: 
https://platform.openai.com/docs/api-reference/chat/object

### Related:

Fixes #486 #492


gehrisandro:
Thanks, @IsraelOrtuno for this PR! 🚀

PR #495, opened 2024-11-06
Moderation Omni - New Categories (Illicit*) support.

iBotPeaches:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

The Omni Moderation models were released, thus deprecating the older models (now referred to as Legacy). This introduced the `illicit` and `illicit/violent` moderation category, but ONLY for Omni. This presented an issue as this library iterated all known types and assumed all models will have it.

Now we only set a Category if we have it and thus support partial support of Omni Moderation.

Ref: https://platform.openai.com/docs/guides/moderation

This does NOT add support for the new `category_applied_input_types` field.


gehrisandro:
Thanks @iBotPeaches for the PR 🥳

PR #502, opened 2024-11-21
feat: add OpenAI compatibility support for Google Gemini

TyperEJ:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

According to the Google Gemini OpenAI compatibility documentation, OpenAI's official Python and Node.js SDKs are supported for compatibility.

Hope that the PHP version can also be updated to reflect this compatibility :)

### Related:

https://ai.google.dev/gemini-api/docs/openai

fixes: #526

### DIFF
--- a/src/Responses/Chat/CreateResponse.php
+++ b/src/Responses/Chat/CreateResponse.php
@@ -12,7 +12,7 @@ use OpenAI\Responses\Meta\MetaInformation;
 use OpenAI\Testing\Responses\Concerns\Fakeable;
 
 /**
- * @implements ResponseContract<array{id: string, object: string, created: int, model: string, system_fingerprint?: string, choices: array<int, array{index: int, message: array{role: string, content: string|null, function_call?: array{name: string, arguments: string}, tool_calls?: array<int, array{id: string, type: string, function: array{name: string, arguments: string}}>}, finish_reason: string|null}>, usage: array{prompt_tokens: int, completion_tokens: int|null, total_tokens: int}}>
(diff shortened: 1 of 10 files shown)


slavkof:
Any chance to get this rolling? :)

npotier:
Hello. Thank you for this PR

any news on it ?

ghostza1209:
hello. any plan to merge this PR?

ghostza1209:
> hello. any plan to merge this PR?
btw, here's my temporary fixed.
https://github.com/openai-php/client/issues/526#issuecomment-2697107494

Capevace:
Would be great to merge this, as this bug currently completely prevents using Gemini models with the SDK! :)

iBotPeaches:
Sorry for delay on this. At the moment there are conflicts, so those would need to be resolved so we can confirm if CI passes, then I can take a look and review this.

iBotPeaches:
It looks good! 

Just a thought about 1 thing. If we have people depending on this package and the `id` property is now nullable - it may be a mini breaking change for them as their static analysis will force them to defend against something that isn't nullable for at least OpenAI.

Do we think thats alright? My only other thought is to dynamically create some random index since it seems Gemini does not if we don't have one in the response. The downside to that is folks may store an ID thinking its from Gemini and its instead some made up value.

iBotPeaches (approved):
I think nullable is right call vs a fake index. It reflects how the official SDKs degrade

iBotPeaches:
thanks!

Capevace:
Thanks so much for the fix! 

Is there any information on the ETA for next version release? We're eagerly awaiting this fix to land! :)

PR #511, opened 2025-01-21
Fix: Support string input for $contents in Exception constructor

sterzuccio:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Fix: Handle cases where $contents is a string instead of an array. Updated constructor to support both array and string types, preventing errors when $contents is not an array.

### DIFF
--- a/src/Exceptions/ErrorException.php
+++ b/src/Exceptions/ErrorException.php
@@ -11,15 +11,21 @@ final class ErrorException extends Exception
     /**
      * Creates a new Exception instance.
      *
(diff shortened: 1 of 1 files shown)


iBotPeaches:
Marking as draft due to CI failures.

iBotPeaches:
This was fixed in https://github.com/openai-php/client/pull/643 as it was refactored for object or string/object calls (to account for audio stringable responses). This shouldn't occur anymore since path of string returns won't have this possibility anymore.

PR #530, opened 2025-03-05
Add compatibility for aliyun LLM APIs

ThorbJ:
### Summary
Some Aliyun API responses do not include the `status_details` field, which causes an error when trying to access it. This PR ensures compatibility by setting `status_details` to `null` when it is missing.

### Changes
- Checks if `status_details` exists in the response before accessing it.
- Defaults `status_details` to `null` if it is not present.
- Ensures compatibility with both OpenAI and Aliyun APIs.

### Why this is needed
Without this fix, Aliyun API responses that lack `status_details` will cause runtime errors, breaking the client integration.

### Testing
- Tested with OpenAI API (unchanged behavior).
- Tested with Aliyun API (no more errors when `status_details` is missing).

### Backward Compatibility
✅ Fully backward compatible. If `status_details` is present, the behavior remains unchanged.

### DIFF
--- a/src/Responses/Files/CreateResponse.php
+++ b/src/Responses/Files/CreateResponse.php
@@ -54,7 +54,7 @@ final class CreateResponse implements ResponseContract, ResponseHasMetaInformati
             $attributes['filename'],
             $attributes['purpose'],
             $attributes['status'],
-            $attributes['status_details'],
+            $attributes['status_details'] ?? null,
             $meta,
         );
     }
--- a/src/Responses/Files/RetrieveResponse.php
+++ b/src/Responses/Files/RetrieveResponse.php
@@ -54,7 +54,7 @@ final class RetrieveResponse implements ResponseContract, ResponseHasMetaInforma
             $attributes['filename'],
             $attributes['purpose'],
             $attributes['status'],
-            $attributes['status_details'],
+            $attributes['status_details'] ?? null,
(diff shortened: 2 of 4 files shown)


iBotPeaches:
Apologies. I can't tell if this passed CI or not since it decayed out. I believe this is fine to merge since the existing types have it as nullable, but I can't confirm till the CI passes.

If you rebase I should be able to confirm.

ThorbJ:
> Apologies. I can't tell if this passed CI or not since it decayed out. I believe this is fine to merge since the existing types have it as nullable, but I can't confirm till the CI passes.
> 
> If you rebase I should be able to confirm.

Hi, I've rebased the branch and marked the PR as ready for review. The workflows are currently awaiting approval since this PR comes from a fork. Could you please approve and run the workflows? Thanks!

iBotPeaches:
looks good. I'm mobile right now but just want to confirm we have a test with this property removed to prevent regressions. It might already exist, but didn't check.

iBotPeaches:
I added a test - just waiting on CI to confirm.

iBotPeaches (approved):
Since `status_details` could be null, I feel safe to hit parity with WebUI/Aliyun to merge this. We just improve our handling to get to the null outcome.

iBotPeaches:
thanks!

Issue #455, opened 2024-07-22
[Bug]: Logprobs property is missing in Chat\CreateResponseChoice


GubaEvgeniy:
### Description

As you can see in the legacy Completion method, support for logprobs is included:
https://github.com/openai-php/client/blob/e9ca2886aa9c0bf60fb3b8df00ad9e5b876a0916/src/Responses/Completions/CreateResponseChoice.php#L12

However, in the current Chat method, support for logprobs is completely missing:
https://github.com/openai-php/client/blob/e9ca2886aa9c0bf60fb3b8df00ad9e5b876a0916/src/Responses/Chat/CreateResponseChoice.php#L7

Please rectify this situation and add support for logprobs to Chat.
Thank you.

### Steps To Reproduce

1. Define a chat request as documented, specifying support for logprobs:
```php
$response = $client->chat()->create([
    'model' => 'gpt-3.5-turbo',
    'messages' => [
        ['role' => 'user', 'content' => 'Hello!'],
    ],
    'logprobs' => true,
]);
```

2. Receive a response from the API that includes a logprobs array in choices.
<img width="320" alt="Снимок экрана 2024-07-22 в 14 52 40" src="https://github.com/user-attachments/assets/564c08be-064e-4b8b-9d2b-bdeaf62cc108">

3. This attribute is missing from the CreateResponseChoice object - https://github.com/openai-php/client/blob/e9ca2886aa9c0bf60fb3b8df00ad9e5b876a0916/src/Responses/Chat/CreateResponseChoice.php#L7

### OpenAI PHP Client Version

0.10.1

### PHP Version

8.2

PR #533, opened 2025-03-11
Chat Response: Add logprobs

klmork:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Add `logprobs` to `CreateResponseChoice`.
- `logprobs` defaults to false if not provided in the request: https://platform.openai.com/docs/api-reference/chat/create#chat-create-logprobs
- We can return the logprobs value or null: https://platform.openai.com/docs/api-reference/chat/object

Updates tests and docs.

This pull does **not:**
- Add `logprobs` to `CreateStreamedResponseChoice`.  
- Add all of the logprob fields at this time (e.g. `refusal` or `content`'s field of `top_logprobs` array).

### Related:

fixes: https://github.com/openai-php/client/issues/455


iBotPeaches:
Do you mind rebasing? I allowed CI to run, but since this is an older PR its still using `cache/v1` which is dead. A simple rebase will solve that and we can see if this will pass CI.

thanks!

klmork:
Thanks @iBotPeaches! Just rebased 👍

iBotPeaches (approved):

iBotPeaches:
thanks! Tested it out a bit with on/off and worked as I expect. Thanks for 1st PR with perfect CI/CD passing and quick response.

klmork:
Awesome! Yay, thanks 😄

Issue #517, opened 2025-02-02
[Bug]: Incompatible with the the DeepSeek R1 model's `reasoning_content` field


charescape:
### Description

![Image](https://github.com/user-attachments/assets/63098ce8-2e15-4b3b-8c9e-4da63124a972)

### Steps To Reproduce

-

### OpenAI PHP Client Version

v0.10.3

### PHP Version

8.3.5

44886:
I solved it！
add this：

![Image](https://github.com/user-attachments/assets/497a4500-13eb-44cf-b463-e8f12ea0c9db)

Snuffel233:
![Image](https://github.com/user-attachments/assets/19a6da5f-0996-47bb-bc33-2ebe8b627041)
It also needs to be modified in the toArray() function, otherwise it will still not work properly
In addition, modifications need to be made accordingly in non-streaming

Cyrille37:
Hi.

`llama-server` has the option `--reasoning-format` that can be `deepseek` to put the reasoning content outside the content, in `message.reasoning_content`

To make my script works fine with reasoning models (like Nemotron-3-Nano-30B-A3B) I have to set option `--reasoning-budget 0` to avoid the reasoning text to be in the message content, which must be valid json for NER extraction.

I understand the tag "#Non-OpenAi Model" but what can we do to distinguish the response vs the reasoning ? And exploiting the `llama-server` option  `--reasoning-format` ?

Thanks for idea & help 💌

iBotPeaches:
If you give me a raw payload response I'll see what I can do. Probably mapping it to the OpenAI format.

Cyrille37:
I'm exploring and coming back in few hours ... 😉

Cyrille37:
No hours needed, `llama-server --reasoning-format deepseek -m Nemotron-3-Nano-30B-A3B-UD-Q8_K_XL.gguf` and changes has https://github.com/openai-php/client/issues/517#issuecomment-2633507230 and https://github.com/openai-php/client/issues/517#issuecomment-2683893826 gives the answer and the reasoning separately : `choices[0]['content']` and `choices[0]['reasoning_content']`.

Here changes I made :

```php
namespace OpenAI\Responses\Chat;
final class CreateResponseMessage
{
    private function __construct(
        public readonly string $role,
        public readonly ?string $content,
        // Cyrille37 https://github.com/openai-php/client/issues/517
        public readonly ?string $reasoning_content,
        public readonly array $annotations,
...
[...]

PR #536, opened 2025-03-13
feat: add reasoning_content field to chat completions response

rxrw:
### What:

- [ ] Bug Fix
- [√] New Feature

### Description:

Add support for the reasoning_content field in chat completions API responses. This field is added at the same level as content in both streaming and non-streaming requests. The field is marked as optional to handle cases where it's not present in the API response.

### Related:

#517

### DIFF
--- a/src/Responses/Chat/CreateResponse.php
+++ b/src/Responses/Chat/CreateResponse.php
@@ -12,12 +12,12 @@ use OpenAI\Responses\Meta\MetaInformation;
 use OpenAI\Testing\Responses\Concerns\Fakeable;
 
 /**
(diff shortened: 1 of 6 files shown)


iBotPeaches:
With the latest merges our pipeline should be runnable again. If you rebase we can see if this change will merge cleanly.

mowangjuanzi:
There is already a PR for the same issue, I think I can close mine.  https://github.com/openai-php/client/pull/536

anhao:
When can this PR be merged

iBotPeaches:
> When can this PR be merged

At the moment its decayed quite a bit with conflicts - so thats probably priority one, then a green pipeline as 2nd. Since its a non-OpenAI change as well it may take some thinking as this property then becomes exposed to all OpenAI implementations without careful filtering logic.

Issue #341, opened 2024-02-22
[Bug]: /vendor/openai-php/client/src/Transporters/HttpTransporter.php(61): json_decode()


chengxs1994:
### Description

When the content returned by openai is not JSON,/vendor/openai php/client/src/Transporters/HttpTransporter. php (61): will throw an exception, but we cannot obtain the specific content by capturing the DeserializableResponse

Could we add the contents attribute to the UnserializableResponse class?

### Steps To Reproduce

`try {
            /** @var array{error?: array{message: string, type: string, code: string}} $data */
            $data = json_decode($contents, true, flags: JSON_THROW_ON_ERROR);
        } catch (JsonException $jsonException) {
            throw new UnserializableResponse($jsonException);
        }`

### OpenAI PHP Client Version

v0.8.1

### PHP Version

8.1

YahorDubrouski:
@chengxs1994 me too

Issue #339, opened 2024-02-19
[Bug]: Uncaught JsonException


sera1t:
### Description

Hello, I installed the library, everything is fine, but for some reason, when I try to make a request, it gives me this error in the console.

[Mon Feb 19 12:51:13 2024] 127.0.0.1:53566 [500]: GET /chat/tet.php - Uncaught JsonException: Syntax error in /var/www/chatbot/vendor/openai-php/client/src/Transporters/HttpTransporter.php:61
Stack trace:
#0 /var/www/chatbot/vendor/openai-php/client/src/Transporters/HttpTransporter.php(61): json_decode()
#1 /var/www/chatbot/vendor/openai-php/client/src/Resources/Chat.php(33): OpenAI\Transporters\HttpTransporter->requestObject()
#2 /var/www/chatbot/public/chat/tet.php(10): OpenAI\Resources\Chat->create()
#3 {main}

Next OpenAI\Exceptions\UnserializableResponse: Syntax error in /var/www/chatbot/vendor/openai-php/client/src/Transporters/HttpTransporter.php:63
Stack trace:
#0 /var/www/chatbot/vendor/openai-php/client/src/Resources/Chat.php(33): OpenAI\Transporters\HttpTransporter->requestObject()
#1 /var/www/chatbot/public/chat/tet.php(10): OpenAI\Resources\Chat->create()
#2 {main}
  thrown in /var/www/chatbot/vendor/openai-php/client/src/Transporters/HttpTransporter.php on line 63

### Steps To Reproduce

I'm trying to run this code 
$client = OpenAI::client($apiKey);

$response = $client->chat()->create([
    'model' => 'gpt-4',
    'messages' => [
        ['role'=>'system', 'content' => 'Assistant'],
        ['role' => 'user', 'content' => 'Hello!'],
    ],
]);

### OpenAI PHP Client Version

v0.8.4

### PHP Version

8.1

PR #440, opened 2024-06-30
Added API error separation by their HTTP codes

LuckThemAll:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Added separation of API errors by their HTTP codes. This enhancement improves error handling by categorizing errors based on their HTTP status codes, allowing for more precise error management and debugging.

### Related:

https://github.com/openai-php/client/issues/402

PR #553, opened 2025-04-10
fix: add throwIfNotSuccessfulStatusCode method to validate HTTP statu…

sinashakouri:
### Summary
This PR introduces stricter HTTP status code validation to the `HttpTransporter` by adding a new method: `throwIfNotSuccessfulStatusCode`. The method ensures that only successful HTTP responses (status codes 200–299) are allowed to proceed. Any other status code results in a thrown `UnexpectedStatusCodeException`.

### Changes
- ✅ Added `throwIfNotSuccessfulStatusCode` method to validate response status codes.
- 🔄 Updated `requestObject`, `requestContent`, and `requestStream` methods to call this validation method.
- 🧪 Added comprehensive tests for various non-successful HTTP status codes.
- 🔧 Fixes issues: #341 and #339.

### Expected Outcome
With this improvement, the client will now provide immediate and clearer feedback when encountering invalid or unexpected HTTP responses, preventing potentially misleading or silent failures.

Looking forward to your feedback!

### DIFF
--- a/src/Exceptions/UnexpectedStatusCodeException.php
+++ b/src/Exceptions/UnexpectedStatusCodeException.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenAI\Exceptions;
+
+use Exception;
+
+final class UnexpectedStatusCodeException extends Exception
+{
+    /**
+     * Creates a new Exception instance.
+     *
+     * @param int $statusCode The HTTP status code.
+     */
+
+    public function __construct(int $statusCode)
+    {
+        $message = "Unexpected status code: {$statusCode}";
+
+        parent::__construct($message, $statusCode);
+    }
+}
--- a/src/Transporters/HttpTransporter.php
+++ b/src/Transporters/HttpTransporter.php
@@ -11,6 +11,7 @@ use OpenAI\Contracts\TransporterContract;
 use OpenAI\Enums\Transporter\ContentType;
 use OpenAI\Exceptions\ErrorException;
 use OpenAI\Exceptions\TransporterException;
+use OpenAI\Exceptions\UnexpectedStatusCodeException;
 use OpenAI\Exceptions\UnserializableResponse;
 use OpenAI\ValueObjects\Transporter\BaseUri;
 use OpenAI\ValueObjects\Transporter\Headers;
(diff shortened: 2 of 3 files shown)


iBotPeaches:
cc @bytestream who has some ideas on Exception improvement.

As I look at this - I'm wondering if we ever get a 200 back from OpenAI that is not valid JSON. There is a common complaint that we don't bubble (in the exceptions) anything about the raw response so folks can't triage the exact reasoning why. If we were to introduce a new Exception I think it should be flexible enough to capture the mixed raw response (could be null or string I'm guessing) so userland can use it - if needed.

However, thats where I was wondering if a HTTP 200 with invalid JSON would arrive, which would throw `UnserializableResponse` that would be missing the raw response content.

Minor @sinashakouri - you just have some CI failures.

bytestream:
@iBotPeaches based on what you [said](https://github.com/openai-php/client/pull/440#issuecomment-2789670989), it sounds like this bits the bill a little better. It just needs to include the request and response objects in the exception so that userland can act accordingly.

> if a HTTP 200 with invalid JSON would arrive

Anything is possible when interacting with third party services. In my mind, it's better to be type certain - PHPStan would not be happy at the moment ;) I'm just not sure whether it's deliberately very loosely typed to allow for:

> this library is used for so many more services than just OpenAI 

I personally thought the clue was in the name `openai-php/client` and it was exclusive to OpenAI 😅

sinashakouri:
Thank you so much for your insightful feedback and suggestions. I really appreciate the time you’ve taken to review the code and provide your valuable insights, which come from your extensive experience.
I’ll definitely take the points you mentioned into consideration and will work on addressing and refining them as suggested.

sinashakouri:
Hey @iBotPeaches, thanks again for the earlier feedback! I’ve made the suggested changes — the exception now includes the request and response, and tests have been updated accordingly. CI is green now as well 🎉

Would appreciate it if you could take another look when you get a chance 🙏

iBotPeaches:
Thanks! I've got my head down trying to finish this Responses API, but will visit this for sure soon

datawench:
Won't this require upstream refactoring, since `VectorStoresFiles::create()` still doesn't catch or throw anything?

bytestream:
The HTTP transporter is what is throwing exceptions based on the response.

f-lombardo:
Any news about this PR?

PR #554, opened 2025-04-11
fix: chat completion choices to allow responses without logprobs field

TyperEJ:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

As it has been some time since I last forked the repository, I noticed from OpenAI's documentation that the chat completion's 'choices' now includes a 'logprobs' field. However, Google Gemini's OpenAI compatibility layer doesn't yet support this field. Therefore, I've modified the code to allow chat completion 'choices' to function without requiring the 'logprobs' field.

### Related:

fixes: #502


iBotPeaches:
thanks!

Issue #500, opened 2024-11-15
[Bug]: No Parameters passable for files()->list()


bytes-commerce:
### Description

The documentation describes how you can pass parameters to `list` calls, and the PHP wrapper has most of it, too.

However, not so much for the files-list. It does not accept parameters which makes querying for pages impossible.

### Steps To Reproduce

```
return $client->vectorStores()->files()->list($vectorStoreId, array_filter([
            'limit' => 100,
            'after' => $after,
        ]));
``` 

This works.

```
return $client->files()->list();
```

This works only without params. Reason is the way its being built:
``` 
public function list(): ListResponse
    {
        $payload = Payload::list('files');

        /** @var Response<array{object: string, data: array<int, array{id: string, object: string, created_at: int, bytes: ?int, filename: string, purpose: string, status: string, status_details: array<array-key, mixed>|string|null}>}> $response */
        $response = $this->transporter->requestObject($payload);

        return ListResponse::from($response->data(), $response->meta());
    }
``` 

Just adapt the method to accept a payload, then pass it to the static method call. 

### OpenAI PHP Client Version

v0.10.1

### PHP Version

8.3

efpapado:
Opening a new PR with same content since the previous one got rejected and the author [wrote](https://github.com/openai-php/client/pull/501#issuecomment-2790944812) that they wouldn't continue working.

PR #557, opened 2025-04-14
Allow arguments to be passed to files list request

assertchris:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

[From a discussion on Twitter...](https://x.com/assertchris/status/1910355384562229390)

TL;DR:
You can't currently pass parameters to the files list request, which means you can only ever get the first 10. This PR adds the ability to paginate, change limit etc.

Unrelated: the refactor commands in readme don't appear to work. Everything else passed.

### Related:

fixes: #500

### DIFF
--- a/src/Contracts/Resources/FilesContract.php
+++ b/src/Contracts/Resources/FilesContract.php
@@ -13,8 +13,10 @@ interface FilesContract
      * Returns a list of files that belong to the user's organization.
      *
      * @see https://platform.openai.com/docs/api-reference/files/list
+     *
+     * @param  array<string, mixed>  $parameters
      */
-    public function list(): ListResponse;
+    public function list(array $parameters = []): ListResponse;
 
     /**
      * Returns information about a specific file.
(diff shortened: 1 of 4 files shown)


iBotPeaches (approved):
good catch. I peeked our tests and didn't see anything that we could augment for this detection in future. The variability of incoming parameters to support changes makes the testing of such a thing not there so not sure we could have caught functions that should have parameters vs not.

+1

nunomaduro:
@iBotPeaches can you add tests on this?

iBotPeaches:
> @iBotPeaches can you add tests on this?

Yeah I'll take a stab at it tomorrow. Trying to think of an elegant approach, maybe spying the transporter to see if it sets what we expect.

iBotPeaches (approved):
Good night of sleep helped. We missed the contract change, thus when applied it made sense where to enhance the test suite. I believe this is good now w/ tests.

iBotPeaches:
thanks!

PR #563, opened 2025-04-23
Chat Support WebsSearch

Saraphoo:
### What:

- [ ] Bug Fix
- [X] New Feature

### Description:

Enabled use of web_search_options for SDK.


iBotPeaches (changes requested):
Looks like you got a bit confused here. The `web_search_options` are an input parameter to the call. We don't generally type those as keep them as `mixed` so changes to OpenAI can be used in this library without any update. If the response changes however - thats where we need to type.

So I can tell you are adding support for the web search aspect, which if you include a relevant model and `web_search_options` as a parameter (unsure if this is required). You get back a response with URL citations of the sources.

```json
[
  {
    "index": 0,
    "message": {
      "role": "assistant",
      "content": "the model response is here...",
      "refusal": null,
      "annotations": [
        {
          "type": "url_citation",
          "url_citation": {
            "end_index": 985,
            "start_index": 764,
            "title": "Page title...",
            "url": "https://..."
          }
        }
      ]
    },
    "finish_reason": "stop"
  }
]
```

It looks like its an array (`annotations`) with 1 type of object `url_citation`. Its possible the configuration picked (`web_search_options`) are reflected back in the response payload, but I'm not seeing that in my testing or the [API docs](https://platform.openai.com/docs/api-reference/chat/object).

Issue #566, opened 2025-04-27
[Bug]: Undefined array key 'created'


the-fermi-paradox:
### Description

Warning: Undefined array key "created" exception is thrown calling 'listModels' on a client with the Gemini URI.

The response from Google doesn't seem to have a 'created' field, which causes a problem when we do:

```php
        public static function from(array $attributes, MetaInformation $meta): self
        {
            return new self(
                $attributes['id'],
                $attributes['object'],
                $attributes['created'],
                $attributes['owned_by'],
                $meta,
            );
        }
```

in `Responses/Models/RetrieveResponse`.

### Steps To Reproduce

```php
$client = \OpenAI::factory()->withApiKey(getenv('GOOGLE_API_KEY'))
    ->withBaseUri('generativelanguage.googleapis.com/v1beta/openai')
    ->make();
$client->models()-list();
```

### OpenAI PHP Client Version

v0.10.3

### PHP Version

8.4

PR #567, opened 2025-04-27
Fix gemini list models

the-fermi-paradox:
### What:

- [x] Bug Fix

### Description:

It seems the 'created' field doesn't always exist on some of these APIs. I experienced the issue with Google Gemini, but after I worked on this code briefly I noticed another pull request existed that was addressing issues with Azure, but by falling back to 0 instead of null. 

This just marks the field as nullable and yields null. I think it's worth discussing (as noted in the linked PR) whether this approach is better or if created should fall back to '0' or something similar.

### Related:

fixes: https://github.com/openai-php/client/issues/566
closes: https://github.com/openai-php/client/pull/458

### DIFF
--- a/src/Responses/Models/ListResponse.php
+++ b/src/Responses/Models/ListResponse.php
@@ -12,12 +12,12 @@ use OpenAI\Responses\Meta\MetaInformation;
 use OpenAI\Testing\Responses\Concerns\Fakeable;
 
 /**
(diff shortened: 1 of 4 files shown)


iBotPeaches (approved):

iBotPeaches:
thanks!

PR #572, opened 2025-04-30
feat(OpenAI): Add category applied input types to moderation response

Mcainson:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This PR adds support for category_applied_input_types to the Moderations endpoint, enabling users to identify which input types (e.g., text, image) were flagged for specific categories in moderation results. 

**Changes**
- Added category_applied_input_types to the Moderations endpoint response structure.
- Updated Moderations::create method to handle and parse category_applied_input_types from the API response.
- Added tests to verify the correct parsing and handling of category_applied_input_types in moderation results.
- Added unit tests for multi-modal (image/text) input handling in Moderations.


abikali:
Nowhere in the documentation https://platform.openai.com/docs/api-reference/moderations/object they mentioned that `category_applied_input_types` is nullable (optional), unless I missed it.

iBotPeaches:
> Nowhere in the documentation https://platform.openai.com/docs/api-reference/moderations/object they mentioned that `category_applied_input_types` is nullable (optional), unless I missed it.

I was running some tests just now. It appears the legacy model - `text-moderation-latest` does not return it, but the new one - `omni-moderation-latest` does.

PR #574, opened 2025-05-02
feat: Support perplexity citations

morloderex:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This adds support for the `citations` attribute that perplexity adds to their responses and the `search_context_size` sub attribute under the usage attribute.

### Related:
No related

### DIFF
--- a/src/Resources/Chat.php
+++ b/src/Resources/Chat.php
@@ -29,7 +29,7 @@ final class Chat implements ChatContract
 
         $payload = Payload::create('chat/completions', $parameters);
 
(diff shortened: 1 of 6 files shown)


morloderex:
@iBotPeaches I haven't seen any other provider that adds these attributes.

The perplexity provider is a bit weird in regards to its response, and it's not probably documented.

iBotPeaches:
@morloderex - Yeah my idea was optional Response classes that are based from the OpenAI one, which then parse custom fields to their provider. Everything else would fallback to existing parsing on OpenAI response class.

morloderex:
@iBotPeaches do you want me to close this pr and open another one with the above in mind then?

iBotPeaches:
@morloderex - Looking for more of a POC or a little pseudo code of how we expect this to work. Generally wondering on an elegant way that allows the library to return a different typed response depending on the provider.

PR #577, opened 2025-05-09
fix(OpenAI): $parameters array is not required in thread 'create' method

erdiarikan:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Threads create requires a $parameters array. It accepts an empty array already. So, we should be able to create threads like this OpenAI::threads()->create() instead of this OpenAI::threads()->create([])


iBotPeaches:
Could you make/adjust the test here to assert you can pass nothing?

https://github.com/openai-php/client/blob/main/tests/Testing/Resources/ThreadsTestResource.php#L9-L20

iBotPeaches:
I went ahead and made the needed changes.

Issue #598, opened 2025-06-11
[Bug]: Undefined array key "owned_by"


absentdream:
### Description

            $client = \OpenAI::factory()
                        ->withBaseUri($endpoint)
                        ->withHttpHeader('api-key', $key)
                        ->withQueryParam('api-version', $version)
                        ->make();
            $response = $client->models()->list();

### Steps To Reproduce

trying to list models on azureAI

### OpenAI PHP Client Version

v0.13.0

### PHP Version

8.2.28

iBotPeaches:
Thanks for the report. I don't have a working Azure connection, so I'll need help with a raw payload. If you make a raw CURL call of that endpoint and provide the response (sanitize what you need). I can take a look

absentdream:
```
{
    "data": [
        {
            "status": "succeeded",
            "capabilities": {
                "fine_tune": false,
                "inference": true,
                "completion": false,
                "chat_completion": false,
                "embeddings": false
            },
            "lifecycle_status": "generally-available",
            "deprecation": {
                "inference": 1751241600
            },
            "id": "dall-e-3-3.0",
            "created_at": 1691712000,
            "object": "model"
        },
        {
            "status": "succeeded",
			........
        }
    ],
    "object": "list"
}
```
models are in the data array.. only showing one as ALOT :)

iBotPeaches:
Okay, OpenAI says that is always required, but also such a tiny payload (`created`, `id`, `object`, `owned_by`). It seems we could make it nullable (at the cost of a tiny breaking change) and fallback to `created_at` to fulfill the `created` prop.

This is Azure OpenAI? Odd they usually have closer parity.

absentdream:
i am fnding there is a standard way... and then a mico$oft way... so many things are 99% the same. but not

absentdream:
and for referance the url

 {{endpoint}}/openai/models?api-version=2025-01-01-preview

and docs used

https://learn.microsoft.com/en-us/rest/api/azureopenai/models/list?view=rest-azureopenai-2024-10-21&tabs=HTTP

iBotPeaches:
Okay I put up a PR that fixes this

PR #599, opened 2025-06-11
fix(Azure): support azure model.list

iBotPeaches:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Adds support for Azure model listing, which uses `created_at` and removes `owned_by`.

### Related:

fixes: #598


iBotPeaches:
Torn on something here. `created_at` is specific to a few other providers. I think adding the `??` fallback to non-standard fields is nice, but our typing says that field doesn't exist and thus fails.

In this PR I added it as an optional field to keep typing happy, but I think I might prefer just suppressing the stan warning.

PR #618, opened 2025-06-24
Provide default image reponse usage values, if they are not returned by the API

svendecabooter:
Provide default image reponse usage values, if they are not returned by the API (e.g. for LiteLLM)

### What:

- [X] Bug Fix
- [ ] New Feature

### Description:

We are using LiteLLM as an AI provider. Supposedly this software would provide the same API as OpenAI.
However, we are encountering issues with image response usage statistics.
The PHP client expects various token usage stats to be returned (such as `total_tokens`, `input_tokens`, ...).
In our default setup of LiteLLM we only retrieve ["total_tokens" => 0], not the other expected data.
Therefore, it would probably be best not to rely on these values always being there, and providing default values in case they are not, instead of crashing.

Other responses seem to work fine for us for now, but we haven't tested all types yet...

### DIFF
--- a/src/Responses/Images/ImageResponseUsage.php
+++ b/src/Responses/Images/ImageResponseUsage.php
@@ -19,10 +19,10 @@ final class ImageResponseUsage
     public static function from(array $attributes): self
     {
         return new self(
-            $attributes['total_tokens'],
-            $attributes['input_tokens'],
-            $attributes['output_tokens'],
-            ImageResponseUsageInputTokensDetails::from($attributes['input_tokens_details']),
+            $attributes['total_tokens'] ?? 0,
+            $attributes['input_tokens'] ?? 0,
+            $attributes['output_tokens'] ?? 0,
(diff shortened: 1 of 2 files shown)


iBotPeaches:
You'll find the problem with just adding null checks is the types don't suggest they are null, so it becomes a bit more of a fix to make things nullable as shown by test failures. In this case falling back to a 0 seems misleading, but making them nullable makes downstream users adapt for a nullable case - so a tough call.

In this case is a `0` or maybe `-1` better?

svendecabooter:
Hmm I'm not familiar enough with this library or the OpenAI API to know what the best approach would be.
In case we set the default value to `-1`, wouldn't the tests also fail then?

Or should we make sure there are always (default) values in the $attributes array earlier on in the code flow?

iBotPeaches:
I'm torn between nullable or `-1` at this point. Nullable because its semantically better, but that causes all downstream users to type for something nullable that generally won't be for OpenAI.

`-1` because it keeps the typing and basically says "I don't know"

PR #619, opened 2025-06-24
Bug/mcp-error-handling

msitarzewski:
test: enhance error handling tests for OutputMcpCall response

### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Added support for handling various MCP call error formats in the OutputMcpCall class. The implementation now properly parses and normalizes error messages returned from OpenAI's MCP calls, regardless of the format they're presented in:
- Direct content arrays with text elements
- Object with nested content arrays
- Simple string errors
- Complex error structures that get JSON-encoded

This ensures that users of the library can consistently access error messages from MCP calls without having to handle different response structures themselves.

### Related:
N/A

### DIFF
--- a/src/Responses/Responses/Output/OutputMcpCall.php
+++ b/src/Responses/Responses/Output/OutputMcpCall.php
@@ -9,7 +9,7 @@ use OpenAI\Responses\Concerns\ArrayAccessible;
 use OpenAI\Testing\Responses\Concerns\Fakeable;
 
 /**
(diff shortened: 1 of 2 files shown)


iBotPeaches:
You might have to rebase your branch. I think you based it off your older branch, prior to the squash merge into main.

iBotPeaches:
Marking as draft in meantime. Conflicting.

PR #624, opened 2025-07-04
Error handling from GrokAI API

joolab:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

GrokAI API returns an error as a string, rather than as an object, as the original OpenAI API does. I propose this adjustment to make the library compatible with the GrokAI API.

### DIFF
--- a/src/Exceptions/ErrorException.php
+++ b/src/Exceptions/ErrorException.php
@@ -11,11 +11,13 @@ final class ErrorException extends Exception
     /**
      * Creates a new Exception instance.
      *
-     * @param  array{message: string|array<int, string>, type: ?string, code: string|int|null}  $contents
+     * @param  string|array{message: string|array<int, string>, type: ?string, code: string|int|null}  $contents
      */
-    public function __construct(private readonly array $contents, private readonly int $statusCode)
+    public function __construct(private readonly string|array $contents, private readonly int $statusCode)
     {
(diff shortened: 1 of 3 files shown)


iBotPeaches:
Do you have a sample of the payload structure or doc link? I haven't used Grok at all yet.

joolab:
> Do you have a sample of the payload structure or doc link? I haven't used Grok at all yet.

They claim that their API is fully compatible with the OpenAI REST API.

https://docs.x.ai/docs/api-reference

iBotPeaches:
Do you have a sample response? What you've changed is like the root level parsing of a response into JSON to make that a string which doesn't seem right. I imagine its still JSON, but with a string on `error`?

The sample will help

joolab:
> Do you have a sample response? What you've changed is like the root level parsing of a response into JSON to make that a string which doesn't seem right. I imagine its still JSON, but with a string on `error`?
> 
> The sample will help

GrokAI:
`{"code":"400","error":"Bad data: Messages cannot be empty"}`

OpenAI:

```
{
  "error": {
    "message": "Invalid 'messages': empty array. Expected an array with minimum length 1, but got an empty array instead.",
    "param": "messages",
    "code": "empty_array"
  }
}
```

iBotPeaches:
Okay thanks. So we can probably look for `code` and `error` at root and detect this as an error. With a change like that and a change to our test suite - we could merge this.

PR #625, opened 2025-07-05
docs: add vision example using image URL

princeaden1:
### What:

- [ ] Bug Fix
- [ ] New Feature
- [x] Docs

### Description:
This pull request adds a new usage example to the README demonstrating how to use the OpenAI to understand and describe images using GPT-4o and an accessible image URL.

###  Example Included

The example shows how to:
- Provide both a text prompt and an image URL
- Output GPT-4o's vision capabilities in a friendly way

###  Why It's Useful
- Showcases advanced GPT-4o usag
- Makes it easy for developers to test image understanding via chat
- Builds on the existing text-only examples by adding multimodal support

### DIFF
--- a/README.md
+++ b/README.md
@@ -499,6 +499,25 @@ $response->usage->completionTokens; // 18,
 $response->usage->totalTokens; // 100
 ```
 
+Creates a chat completion with image input via `image_url`.  
+Useful for describing and analyzing visual content.
+
+```php
+$response = $client->chat()->create([
+    'model' => 'gpt-4o',
+    'messages' => [
+        [
+            'role' => 'user',
+            'content' => [
+                ['type' => 'text', 'text' => 'What is in this image?'],
+                // Replace with a real, accessible image
+                ['type' => 'image_url', 'image_url' => ['url' => 'https://example.com/image.jpg']], 
+            ]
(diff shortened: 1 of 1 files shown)


iBotPeaches (changes requested):
I think you misunderstood the docs a bit - since you made a new top level "Vision" that is misleading because this is just a regular ole Chat creation endpoint with `image_url` support. So "Vision resource" suggests there is a new top level Vision API set of endpoints which there isn't - this is just a parameter on Chat.

The amount of variables is a bit heavy for docs. Since we have to balance like 200 endpoints its all about short/sweet. Take a look at the Chat example - https://github.com/openai-php/client?tab=readme-ov-file#chat-resource and see how we have examples for function calls & tool calls - you could expand that with an image_url example.

However saying that too - why increase our documentation for Chat? Responses is the future.

princeaden1:
Thank you for the feedback,  I've force pushed a clean version of the branch based on main, with a single commit.  The image input example is now added under the Chat section, following the existing structure and style. Let me know if anything else needs adjustment. Thank you.

iBotPeaches (approved):

iBotPeaches:
thanks!

Issue #620, opened 2025-06-25
Is there suport for Update vector store file attributes request?


kolyabres:
https://platform.openai.com/docs/api-reference/vector-stores-files/updateAttributes

iBotPeaches:
Does appear to be missing - https://github.com/openai-php/client/blob/main/src/Resources/VectorStoresFiles.php

Good call out.

PR #626, opened 2025-07-06
feat(OpenAI): add updateAttributes method to VectorStoresFiles

Bombaninha:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Implements the `updateAttributes` method for Vector Store Files to allow updating file attributes via the OpenAI API. This method enables users to modify the key-value pairs attached to vector store files.

Changes include:
- Added `updateAttributes` method to `VectorStoresFiles` class
- Added test for the new method

### Related:

#620

### DIFF
--- a/src/Resources/VectorStoresFiles.php
+++ b/src/Resources/VectorStoresFiles.php
@@ -65,6 +65,23 @@ final class VectorStoresFiles implements VectorStoresFilesContract
     }
 
     /**
+     * Update attributes on a vector store file.
+     *
+     * @see https://platform.openai.com/docs/api-reference/vector-stores-files/updateFile
+     *
+     * @param  array<string, mixed>  $parameters
+     */
+    public function updateAttributes(string $vectorStoreId, string $fileId, array $parameters): VectorStoreFileResponse
+    {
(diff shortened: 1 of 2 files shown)


src/Resources/VectorStoresFiles.php
```diff
@@ -64,6 +64,23 @@ public function retrieve(string $vectorStoreId, string $fileId): VectorStoreFile
         return VectorStoreFileResponse::from($response->data(), $response->meta());
     }
 
+    /**
+     * Update attributes on a vector store file.
+     *
+     * @see https://platform.openai.com/docs/api-reference/vector-stores-files/updateFile
```
iBotPeaches, line 70: This link is not working for me, I searched and found this one - https://platform.openai.com/docs/api-reference/vector-stores-files/updateAttributes to confirm. Otherwise looks good.

Bombaninha: You're right.
I confused it with the deleteFile endpoint. Thanks for pointing it out! 
I've already pushed the fix.

iBotPeaches:
thanks!

PR #634, opened 2025-07-22
Add model field to CreateResponse for embeddings

bemtorres:
### What:

- [ ] Bug Fix
- [x ] New Feature

### Description:

This pull request adds support for the model field in the CreateResponse class for embeddings.

The OpenAI API's embeddings response includes a model property, but it wasn’t being exposed in the PHP client. This update fixes that by adding the model field to the response class, so users can now access it directly—just like the API intends.

- Added the model property to the CreateResponse class.
- Updated the constructor, from method, and toArray method to handle the model field.
- Adjusted related tests to check for the presence of the model property.

https://platform.openai.com/docs/api-reference/embeddings/create

reposnse 

```
{
  "object": "list",
  "data": [
    {
      "object": "embedding",
      "embedding": [
        0.0023064255,
        -0.009327292,
        .... (1536 floats total for ada-002)
        -0.0028842222,
      ],
      "index": 0
    }
  ],
  "model": "text-embedding-ada-002",
  "usage": {
    "prompt_tokens": 8,
    "total_tokens": 8
  }
}

```

### DIFF
--- a/src/Responses/Embeddings/CreateResponse.php
+++ b/src/Responses/Embeddings/CreateResponse.php
@@ -29,6 +29,7 @@ final class CreateResponse implements ResponseContract, ResponseHasMetaInformati
      */
     private function __construct(
         public readonly string $object,
+        public readonly string $model,
         public readonly array $embeddings,
         public readonly ?CreateResponseUsage $usage,
         private readonly MetaInformation $meta,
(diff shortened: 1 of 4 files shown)


iBotPeaches:
thanks! If you remove inline comments and add to doc block this should be good.

iBotPeaches (changes requested):
This property needs to be added as nullable (or missing) as parity with the other providers don't return this.

Then should be good to go

iBotPeaches (approved):

iBotPeaches:
thanks!

Issue #549, opened 2025-04-08
[Bug]: PHP Fatal error: Uncaught TypeError


smazur:
### Description

```
PHP Fatal error: Uncaught TypeError: OpenAI\Responses\Threads\Messages\ThreadMessageResponse::from(): Argument #1 ($attributes) must be of type array, string given, called in .../vendor/openai-php/client/src/Resources/ThreadsMessages.php on line 32
```

### Steps To Reproduce

This error happens irregularly during following call chain:

```php
$client->threads()->messages()->create( $thread_id, $message );
```

### OpenAI PHP Client Version

v0.10.3

### PHP Version

8.2.16

### Notes

After error_log debugging I found that API replied with 'Bad Gateway' response and 'Content-Type: text/plain`.

iBotPeaches:
Yeah generally at the moment if the response isn't JSON - the handling isn't the best. This has manifested differently depending on the call. Some others related:

 * https://github.com/openai-php/client/issues/341
 * https://github.com/openai-php/client/issues/339

bytestream:
Also https://github.com/openai-php/client/issues/478

iBotPeaches:
We have a fix in progress to rework a bit of the exception handling for that use-case. Trying to gather more samples in the field of the odd OpenAI issues to confirm: https://github.com/openai-php/client/pull/553

Issue #402, opened 2024-05-24
Exceptions should have more detailed information, like what was the RESPONSE text.


pavelthq:
https://github.com/openai-php/client/blob/7b5652b5e107b04b67aa5bc6560acf162ef2be97/src/Transporters/HttpTransporter.php#L63

JsonException: Syntax error in /var/www/html/chatgpt/vendor/openai-php/client/src/Transporters/HttpTransporter.php:61

Next OpenAI\Exceptions\UnserializableResponse: Syntax error in /var/www/html/chatgpt/vendor/openai-php/client/src/Transporters/HttpTransporter.php:63
Stack trace:
#0 /var/www/html/chatgpt/vendor/openai-php/client/src/Resources/Chat.php(33): OpenAI\Transporters\HttpTransporter->requestObject()
#1 /var/www/html/chatgpt/app/Jobs/RequestItemGpt.php(159): OpenAI\Resources\Chat->create()

gehrisandro:
@pavelthq Good idea. Could you make a PR?

Radiergummi:
How about adding the request and response instances to the exception?

Issue #326, opened 2024-01-26
[Request] I want to get raw response headers.


rkgo55:
Thanks again.
This library has a parsing function in MetaInformation.php, but this parsing is not sufficient when using services other than OpenAI.

Example
1) The following headers are added only when using Azure API Management.
```
  "x-openai-backendurl" => array:1 [
    0 => "https://{DOMAIN}/openai/deployments/{MODEL_NAME}"
  ]
```

It seems very difficult to cover everything in the library, so we would like the raw response headers to be included in MetaInformation.php

gehrisandro:
Hi @rkgo55 

Thank you very much for your suggestion. We may add this in a future release.

Issue #59, opened 2023-03-01
add new audio transcription and translation endpoints (API version 1.2.0)


gehrisandro:
https://platform.openai.com/docs/api-reference/audio

gehrisandro:
This seems to be fun 🙈 

The request has a parameter `response_format` which can be set to one of the following: json, text, srt, verbose_json, or vtt.

As the parameter name suggests, this results in very different return messages. Here are example messages of the various formats:

**text**
`Hello, how are you? Fine and you? What is your name? My name is Tom.\n`

**json**
`{"text":"Hello, how are you? Fine and you? What is your name? My name is Tom."}`

**verbose_json**
```json
{
  "task": "transcribe",
  "language": "english",
  "duration": 11.83,
  "segments": [
    {
      "id": 0,
      "seek": 0,
      "start": 0.0,
      "end": 12.32,
      "text": " Hello, how are you? Fine and you? What is your name? My name is Tom.",
      "tokens": [
        50364,
        ...,
        50980
      ],
      "temperature": 0.0,
      "avg_logprob": -0.3095785223919412,
      "compression_ratio": 1.0625,
      "no_speech_prob": 0.031427089124917984,
      "transient": false
    }
  ],
  "text": "Hello, how are you? Fine and you? What is your name? My name is Tom."
}
```

**vtt**
```
WEBVTT\n
\n
00:00:00.000 --> 00:00:12.320\n
Hello, how are you? Fine and you? What is your name? My name is Tom.\n
\n
```

**srt**
```
1\n
00:00:00,000 --> 00:00:12,320\n
Hello, how are you? Fine and you? What is your name? My name is Tom.\n
\n
\n
```

I am not an audio expert but I guess there is a good reason to have different out formats. But in a REST way this is quiet weird, as it breaks with every REST standard.

Right now I am not sure how to handle this properly. Currently I have the following ideas:

1) Always force `verbose_json` and return a corresponding response which has more or less the same format as all the other responses.

2) Let the API user control the `response_format` parameter and return a response corresponding to the format the user has chosen. `text`=> `TranscriptionTextResponse`; `vtt`=> `TranscriptionVttResponse` etc.

[...]

PR #643, opened 2025-08-14
chore(meta): rework exception handling

iBotPeaches:
### What:

- [x] Bug Fix
- [x] New Feature

### Description:
People have been complaining about phantom issues for awhile that parsers get strings, but expect objects. This has been reported a lot of times.

Folks have tried to fix this many times by either overhauling exceptions, detecting http status codes, introducing many new exceptions and all of this is not addressing the root problem.

A long time ago an issue was created to support audio transcription api (https://github.com/openai-php/client/issues/59) and that endpoint has a powerhouse of a feature.

> Select the format of the output, in one of these options: json, text, srt, verbose_json, or vtt

The endpoint can change its return to a variety of types. So base Response classes were adapted to support string/array to handle all of those. The problem however that every single other endpoint only supports json/array. So now that the base class supports a string (the other types) they can crash in odd situations - like when get an internal error (string) that is then expected to be handled by every single entity, when only Audio is prepped for it.

You'll probably get a 404 not found in pure text from some OpenAI load balancer at some point. At thousands of OpenAI calls a day we get roughly 0-3 of these errors every single day. Its an ugly error - its not a structured OpenAI return. Since the base classes support strings after #59 - it moves onward to any other non-audio class and crashes. This exception is ugly

> Argument #1 ($contents) must be of type array, string given

Been reported for years. So I'm on a journey to resolve this at the root.

### Steps
 - [x] Remove string as a valid type for Base Response
 - [x] Introduce Audio Base Response (string/array)
 - [x] Augment real life tests in my test repo for all different formats for audio/transcription to prevent regressions (https://github.com/iBotPeaches/openai-php-laravel-test/commit/a9be8878112b67bb990d7a9277536102fddcc232)
 - [x] Add ResponseInterface to exceptions to catch raw response easier
 - [x] Correct Audio class with all these changes
 - [x] Tests for sampled 404 issue
 - [x] Tests for Response included in exceptions
 - [x] Capture unknown headers in a new meta class
 - [ ] (Stretch) Add a PSR-18 Laravel HTTP/Client implementation into Laravel repo to aid in tracing these calls.

### Related:

fixes https://github.com/openai-php/client/issues/326
fixes https://github.com/openai-php/client/issues/339
fixes https://github.com/openai-php/client/issues/341
fixes https://github.com/openai-php/client/issues/402
fixes https://github.com/openai-php/client/issues/549

closes https://github.com/openai-php/client/pull/553
closes https://github.com/openai-php/client/pull/440
closes https://github.com/openai-php/client/pull/429
closes https://github.com/openai-php/client/pull/359

### DIFF
--- a/src/Contracts/TransporterContract.php
+++ b/src/Contracts/TransporterContract.php
@@ -7,6 +7,7 @@ namespace OpenAI\Contracts;
 use OpenAI\Exceptions\ErrorException;
 use OpenAI\Exceptions\TransporterException;
 use OpenAI\Exceptions\UnserializableResponse;
+use OpenAI\ValueObjects\Transporter\AdaptableResponse;
 use OpenAI\ValueObjects\Transporter\Payload;
 use OpenAI\ValueObjects\Transporter\Response;
 use Psr\Http\Message\ResponseInterface;
@@ -17,7 +18,7 @@ use Psr\Http\Message\ResponseInterface;
 interface TransporterContract
 {
     /**
-     * Sends a request to a server.
+     * Sends a request to a server expecting an object back.
      *
      * @return Response<array<array-key, mixed>>
      *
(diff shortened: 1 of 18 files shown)


iBotPeaches:
thanks @bytestream! pinged a few coworkers to double check the work as well.

bytestream:
@iBotPeaches I've just noticed that you can't access the headers on rate limited requests? I think `$response` needs to be included in the `throw new ErrorException` calls - like you did in the other cases

iBotPeaches:
> @iBotPeaches I've just noticed that you can't access the headers on rate limited requests? I think `$response` needs to be included in the `throw new ErrorException` calls - like you did in the other cases

good catch. I'll do that today/tomorrow

Issue #581, opened 2025-05-21
[Bug]: Fakeable trait incorrectly merges array properties causing contaminated test responses


bartsome:
### Description

When I was using CreateResponse::fake() to create test fixtures for response api, I've noticed the Fakeable trait's buildAttributes method improperly merges indexed arrays like output from the default fixtures with the provided override data, resulting in unexpected content in test responses.

### Steps To Reproduce
```php
// Create a fake response with specific output
$response = CreateResponse::fake([
    'output' => [
        [
            'type' => 'message',
            'id' => 'msg_1',
            'role' => 'assistant',
            'status' => 'completed',
            'content' => [
                [
                    'type' => 'output_text',
                    'text' => 'This is a test response',
                    'annotations' => [],
                ],
            ],
        ],
    ],
]);

// Actual output contains both the test data AND data from CreateResponseFixture.php
// Unexpected output like: "This is a test responseAs of today, March 9, 2025, one notable positive news story..."
```
### OpenAI PHP Client Version

v0.13.0

### PHP Version

8.4.0

### Notes

Cursor suggested solution:
```php
private static function buildAttributes(array $original, array $override): array
{
    $new = [];

    foreach ($original as $key => $entry) {
        // Check if this is an array that should be replaced entirely rather than merged
        if (isset($override[$key]) && is_array($entry) && is_array($override[$key]) && 
            (in_array($key, ['output', 'tools', 'annotations']) || isset($override[$key][0]))) {
            // For arrays of objects (indexed arrays), replace entirely instead of merging
            $new[$key] = $override[$key];
        } else {
            // For other arrays, continue with recursive merging
            $new[$key] = is_array($entry) ?
                self::buildAttributes($entry, $override[$key] ?? []) :
                $override[$key] ?? $entry;
        }
        unset($override[$key]);
    }

    // Add any remaining overrides
    foreach ($override as $key => $value) {
        $new[$key] = $value;
    }

    return $new;
}

```

iBotPeaches:
Interesting. So we need to write a test to test our testing code.

iBotPeaches:
I took a look at this and there is no good fix for this. 14 places in the codebase make the assumption to merge nested arrays. Now I agree with you - it's an odd experience, but plucking out attributes to build a whitelist of attributes to ignore isn't exactly a good fix either.

`fake()` might need to expand with some method to dictate merge/replace on nested attributes.

https://github.com/openai-php/client/pull/593

droath:
[...]

PR #646, opened 2025-08-14
#581: Add `OverrideStrategy` enum and enable merge/replace 

droath:
### What:

- [ X] Bug Fix
- [ ] New Feature

### Description:

I ran into an issue that relates to the bug described at https://github.com/openai‑php/client/issues/581. This PR fixes that issue, and all PHPUnit tests have passed with these changes.

### DIFF
--- a/src/Testing/Enums/OverrideStrategy.php
+++ b/src/Testing/Enums/OverrideStrategy.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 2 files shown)


droath:
Okay—I was able to get the unit tests and linting all passing locally. I've added a test case that will ensure it works as expected when using either the replace or merge strategy.

iBotPeaches (approved):
Looks great! thanks.

PR #658, opened 2025-08-26
feat(OpenAI): add Conversation API

iBotPeaches:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:
Adds Conversation API and Conversation Item API

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,10 @@ If you or your business relies on this package, it's important to support the de
 - [Usage](#usage)
   - [Models Resource](#models-resource)
   - [Responses Resource](#responses-resource)
+  - [Responses Conversations Resource](#responses-conversations-resource)
+  - [Responses Conversations Items Resource](#responses-conversations-items-resource)
   - [Containers Resource](#containers-resource)
-  - [Containers Files Resource](#container-files-resource)
+  - [Containers Files Resource](#containers-files-resource)
   - [Chat Resource](#chat-resource)
   - [Audio Resource](#audio-resource)
   - [Embeddings Resource](#embeddings-resource)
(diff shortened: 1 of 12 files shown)


iBotPeaches:
I originally designed this under Response API, because this is all its used for at present, but the URL structure of the endpoints does not sit under `/responses` like say Vector Files or Container Files. So now I believe its a new top level entity, but all its payload/responses are chunks of Response API.

mydnic:
> I originally designed this under Response API, because this is all its used for at present, but the URL structure of the endpoints does not sit under `/responses` like say Vector Files or Container Files. So now I believe its a new top level entity, but all its payload/responses are chunks of Response API.

Isn't this the replacement of the old Threads API ? I believe the Conversations API is required to be able to migrate from Assistants/Threads/Runs API to the Response API

iBotPeaches:
> Isn't this the replacement of the old Threads API ? I believe the Conversations API is required to be able to migrate from Assistants/Threads/Runs API to the Response API

Screenshot From 2025-09-01 06-38-35

Yes thats correct.

iBotPeaches:
There is a lot of overlap, but not perfect with Response API. Once I peeked official JS SDK I noticed.

```js
/**
 * A single item within a conversation. The set of possible types are the same as
 * the `output` type of a
 * [Response object](https://platform.openai.com/docs/api-reference/responses/object#responses/object-output).
 */
export type ConversationItem =
  | ConversationsAPI.Message
  | ResponsesAPI.ResponseFunctionToolCallItem
  | ResponsesAPI.ResponseFunctionToolCallOutputItem
  | ResponsesAPI.ResponseFileSearchToolCall
  | ResponsesAPI.ResponseFunctionWebSearch
  | ConversationItem.ImageGenerationCall
  | ResponsesAPI.ResponseComputerToolCall
  | ResponsesAPI.ResponseComputerToolCallOutputItem
  | ResponsesAPI.ResponseReasoningItem
  | ResponsesAPI.ResponseCodeInterpreterToolCall
  | ConversationItem.LocalShellCall
  | ConversationItem.LocalShellCallOutput
  | ConversationItem.McpListTools
  | ConversationItem.McpApprovalRequest
  | ConversationItem.McpApprovalResponse
  | ConversationItem.McpCall
  | ResponsesAPI.ResponseCustomToolCall
  | ResponsesAPI.ResponseCustomToolCallOutput;
```

So my changes to the Response InputMessage is invalid as that does not stay true for the Response InputMessages, my indicator is all the classes the official SDK duplicated above instead of re-using the Response API iteration. There must be a reason for each one.

I'll revert changes for Message since I can't unify Response/Conversation into one singular message class. I wish I could, but noticing how official SDK duplicates so many types they must fear the same problem - tiny breaking changes.

iBotPeaches:
Okay I believe is done now. I'll do some code review on it to double check, but I believe after battle testing it with my own little [scripts](https://github.com/iBotPeaches/openai-php-laravel-test/blob/master/app/Console/Commands/ConversationsTest.php) - its ready.

In this journey I found a few new properties, events in Responses that I need to port over. I'll do those in another PR. This brings in a lot of support for Local Shell & Custom Tool which were not yet typed in Responses or Conversation.

mydnic:
@iBotPeaches I was trying to upgrade my project to migrate from Assistants to Prompts + Conversations by using your latest changes

I got this error however

```
OpenAI::client(config('services.openai.key'))->conversations()->create()
```

> OpenAI\Exceptions\ErrorException  Unknown beta requested: 'assistants'.

iBotPeaches:
> > OpenAI\Exceptions\ErrorException  Unknown beta requested: 'assistants'.

Hmm that would be problematic - since that header is applied by default here - https://github.com/openai-php/client/blob/main/src/OpenAI.php#L19

Not sure why I'm not replicating, but I'll dig into this

mydnic:
> > > OpenAI\Exceptions\ErrorException  Unknown beta requested: 'assistants'.
> 
> Hmm that would be problematic - since that header is applied by default here - https://github.com/openai-php/client/blob/main/src/OpenAI.php#L19
> 
> Not sure why I'm not replicating, but I'll dig into this

yeah when I comment this line, it works.

iBotPeaches:
> yeah when I comment this line, it works.

thanks. Still researching, but started this - https://github.com/openai-php/client/pull/688

Issue #660, opened 2025-08-26
[Bug]: Responses API: inconsistent MCP tool error shape causes Undefined array key "code"


e4se:
### Description

Hi everybody,

When using the Responses API with an MCP Server, the error field shape differs depending on where the failure occurs.

For transport failures, the error is an http_error object: { code, message }.

For tool execution failures, the error is an mcp_tool_execution_error object that does not include code.
This inconsistency leads to: GenericResponseError: Undefined array key "code" in clients that assume error.code always exists.

Examples:
Transport/HTTP error (works as expected):
```
{
  "type": "http_error",
  "error": { "code": 500, "message": "Internal Server Error" }
}
```

Tool execution error (no code, causes exception):
```
{
  "type": "mcp_tool_execution_error",
  "content": [
    {
      "type": "text",
      "text": "Unknown error",
      "annotations": null,
      "meta": null
    }
  ]
}
```

The tool-execution error shape appears to follow the MCP spec:
https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling

### Steps To Reproduce

Attach an MCP server with a tool that deliberately throws during execution to the Responses API, prompt the model to call that tool, and observe the response returns error: { "type": "mcp_tool_execution_error", "content": [...] } (no code/message), triggering a Undefined array key "code"

### OpenAI PHP Client Version

v0.16.0

### PHP Version

8.4

iBotPeaches:
Guess I need to find an easy/public MCP that throws errors. Here is my previous sample I used to design the other MCP support - https://github.com/iBotPeaches/openai-php-laravel-test/blob/master/app/Console/Commands/ResponsesMcpTest.php

e4se:
> Guess I need to find an easy/public MCP that throws errors. Here is my previous sample I used to design the other MCP support - https://github.com/iBotPeaches/openai-php-laravel-test/blob/master/app/Console/Commands/ResponsesMcpTest.php

For example: 
{
    server_label: 'deploy-html',
    server_url: 'https://remote.mcpservers.org/edgeone-pages/mcp'
}

Test message: "Use ur method deploy-html to deploy this text "test""

iBotPeaches:
```
   ErrorException 

  Undefined array key "code"

  at openai-client/src/Responses/Responses/GenericResponseError.php:36
     32▕      */
     33▕     public static function from(array $attributes): self
     34▕     {
     35▕         return new self(
  ➜  36▕             code: (string) $attributes['code'],
[...]

PR #661, opened 2025-08-27
fix(OpenAI): support mcp error objects in response api

iBotPeaches:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

MCP is messy. OpenAI presently uses `error` tag in 3 different ways.

 1. A basic string
 2. An array of `code` and `message` (OpenAI structured error)
 3. An array of `type` and `content` (mcp structured error)
 
It would seem odd for end user to be responsible for 3 different error returns, but I didn't know 2/3 of these formats existed until after first implementation was written. To reduce shift in userland code we map all exceptions back to a Generic exception. This seems like best path until we have more results and usage as Response API gains adoption to figure out how to refine further.

### Related:

fixes: #660

### DIFF
--- a/src/Responses/Responses/GenericResponseError.php
+++ b/src/Responses/Responses/GenericResponseError.php
@@ -9,11 +9,11 @@ use OpenAI\Responses\Concerns\ArrayAccessible;
 use OpenAI\Testing\Responses\Concerns\Fakeable;
 
 /**
- * @phpstan-type ErrorType array{code: string|int, message: string}
+ * @phpstan-type ErrorType array{code?: string|int, message?: string}
  *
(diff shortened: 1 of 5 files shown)


e4se:
I don’t think it makes sense to extend McpGenericResponseError from GenericResponseError. For that reason, renaming type to code and hardcoding the message doesn’t seem ideal either.

It would be cleaner to simply create a new class — McpResponseError — with the properties type and content.
It doesn’t break anything, since GenericResponseError only appeared in last two days.

iBotPeaches:
Thanks for feedback. I have reworked a bit to have it standalone.

PR #662, opened 2025-08-27
Handle optional attributes in CreateResponse

louisbels:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

 Fix compatibility with Scaleway AI API responses by making store and text fields optional in CreateResponse.

  The Scaleway AI API (which uses OpenAI-compatible endpoints) doesn't always include the store and text fields in their response format, causing "Undefined array key" errors when using the OpenAI PHP client.

  Changes made:
  - Make store field optional with default value true when missing
  - Make text field optional and nullable when missing
  - Update toArray() method to handle nullable text field

  This maintains backward compatibility with OpenAI's API while allowing the client to work with other OpenAI-compatible APIs like Scaleway.

### Related:

 Description:

  Fix compatibility with Scaleway AI API by making store and text fields optional in CreateResponse.

  Scaleway's OpenAI-compatible API doesn't always include these fields, causing "Undefined array key" errors. This change adds proper fallbacks while maintaining backward compatibility with OpenAI's API.


iBotPeaches:
I triggered CI pipeline. Since you made some fields nullable, etc. The typing will need changes and I'd prefer a test where you null out these fields to assert it works as you expect.

PR #665, opened 2025-08-28
feat: hosted-tool-choice-web-search

ymktmk:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

- Support for the web_search tool in the Responses API
    - Added web_search to the match branch of the tool/tool_choice in CreateResponse/RetrieveResponse
    - Added web_search to the type annotations of WebSearchTool/HostedToolChoice
- This prevents runtime exceptions (unhandled match cases) when the API returns web_search
- There is no change in previous behavior; only the accepted values ​​have been expanded (compatible).


iBotPeaches (commented):
> The type of the web search tool. One of `web_search` or `web_search_2025_08_26`.

Should you add the other one as well? Just checked docs to validate this and saw another.

iBotPeaches (approved):
We an iterate later on more support. This will add the more urgent need for now.

PR #689, opened 2025-09-24
Fix undefined array key errors for xAI Responses API

dnnp2011:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

xAI Responses API does not return certain attributes. The `Responses/Responses/CreateResponse` class already allows these fields to be null.

- Added null fallbacks to `instructions, and `truncation` attributes to avoid errors when handling xAI Responses API response _(e.g. grok-4)_

See [xAI Docs](https://docs.x.ai/docs/api-reference#create-new-response)


dnnp2011:
Yeah I noticed that after I opened it last night, I'll fix it today to resolve the failing tests

iBotPeaches:
thanks for quick fixes

PR #691, opened 2025-09-28
Update README.md

carvemerson:
### What:

- [ ] Bug Fix
- [ ] New Feature
- [x] README update 

### Description:

Fix tool call example on README

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,50 @@ $response->usage->totalTokens; // 123
 $response->toArray(); // ['id' => 'resp_67ccd2bed1ec8190b14f964abc054267', ...]
 ```
 
+Create a model response with a function tool.
+
+```php
+$response = $client->responses()->create([
+    'model' => 'gpt-4o-mini',
+    'tools' => [
+        [
+            'type' => 'function',
+            'name' => 'get_temperature',
+            'description' => 'Get the current temperature in a given location',
+            'parameters' => [
+                'type' => 'object',
+                'properties' => [
+                    'location' => [
+                        'type' => 'string',
+                        'description' => 'The city and state, e.g. San Francisco, CA',
+                    ],
+                    'unit' => [
+                        'type' => 'string',
+                        'enum' => ['celsius', 'fahrenheit'],
+                    ],
+                ],
+                'required' => ['location'],
+            ],
+        ]
+    ],
+    'input' => "What is the temperature in Rio Grande do Norte, Brazil?",
+]);
+
+foreach ($response->output as $item) {
+    if ($item->type === 'function_call') {
+        $name = $item->name ?? null;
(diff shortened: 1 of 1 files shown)


iBotPeaches (changes requested):
I'll admit I haven't used Chat Completion since Response API came out, but this seems wrong. So can you provide why you think this needs to be different? Is it the case that OpenAI docs are wrong?

I'm looking at this [sample](https://platform.openai.com/docs/api-reference/chat/create)

```bash
curl https://api.openai.com/v1/chat/completions \
-H "Content-Type: application/json" \
-H "Authorization: Bearer $OPENAI_API_KEY" \
-d '{
  "model": "gpt-4.1",
  "messages": [
    {
      "role": "user",
      "content": "What is the weather like in Boston today?"
    }
  ],
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "get_current_weather",
        "description": "Get the current weather in a given location",
        "parameters": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string",
              "description": "The city and state, e.g. San Francisco, CA"
            },
            "unit": {
              "type": "string",
              "enum": ["celsius", "fahrenheit"]
            }
          },
          "required": ["location"]
        }
      }
    }
  ],
  "tool_choice": "auto"
}'
```

thomaspapanikolaou:
There are differences in the tools schema when moving to the Responses API, acknowledge d here

[https://platform.openai.com/docs/guides/migrate-to-responses](url)

Specifically, the new schema is shown here

[https://platform.openai.com/docs/guides/function-calling](url)

So the README is correct for the Chat Resource. You need a different schema for the Responses resource (no example provided today, except the built-in `web_search_preview` tool).

I hope this helps.

iBotPeaches:
Okay so that was my confusion. You wanted to add an example for function call for Responses API, but edited the Chat completion one on accident?

thomaspapanikolaou:
I guess that's what Emerson (the originator) meant. This is Thomas here Connor, trying to help.

carvemerson:
Sorry for the confusion. My goal was to provide an example of how to use the function tool, but I didn't noticed that it was on the wrong place. I’ve reverted my earlier change and added a new section with the example. Hopefully, this will be useful for other developers.

README.md
```diff
@@ -312,6 +312,55 @@ $response->hasMore; // false
 $response->toArray(); // ['object' => 'list', 'data' => [...], ...]
 ```
 
+#### `function tool`
```
iBotPeaches, line 315: Can we remove the header (since it makes it seem like its a top-level command) and do a simple line beneath the create call?

Image

Much like we did in Chat

carvemerson:
Thx for the feedback, I did the change.

iBotPeaches:
thanks!

Issue #695, opened 2025-10-02
[Bug]: Missing `filename` property on `OutputMessageContentOutputTextAnnotationsFileCitation`


qwertyquest:
### Description

OpenAI returns the following message response for file citations:

```
{
  "id": "msg_0c[...]",
  "type": "message",
  "role": "assistant",
  "status": "completed",
  "content": [
    {
      "type": "output_text",
      "annotations": [
        {
          "type": "file_citation",
          "file_id": "file-9bD[...]",
          "filename": "test-file.txt",
          "index": 138
        },
        {
          "type": "file_citation",
          "file_id": "file-9bD[...]",
          "filename": "test-file.txt",
          "index": 294
        }
      ],
      "text": "[...]"
    }
  ]
}
```

[OutputMessageContentOutputTextAnnotationsFileCitation](https://github.com/openai-php/client/blob/main/src/Responses/Responses/Output/OutputMessageContentOutputTextAnnotationsFileCitation.php) appears to be missing the `filename` property, striping it from the final response.

### Steps To Reproduce

n/a

### OpenAI PHP Client Version

v0.17.0

### PHP Version

8.2.5

PR #696, opened 2025-10-03
Add `filename` property to OutputMessageContentOutputTextAnnotationsF…

qwertyquest:
### What:

- [X] Bug Fix
- [ ] New Feature

### Description:

Adds missing `filename` property to OutputMessageContentOutputTextAnnotationsFileCitation.php

### Related:

Fixes [#695](https://github.com/openai-php/client/issues/695)

### DIFF
--- a/src/Responses/Responses/Output/OutputMessageContentOutputTextAnnotationsFileCitation.php
+++ b/src/Responses/Responses/Output/OutputMessageContentOutputTextAnnotationsFileCitation.php
@@ -9,14 +9,14 @@ use OpenAI\Responses\Concerns\ArrayAccessible;
 use OpenAI\Testing\Responses\Concerns\Fakeable;
 
 /**
- * @phpstan-type FileCitationType array{file_id: string, index: int, type: 'file_citation'}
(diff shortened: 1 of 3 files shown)


iBotPeaches (commented):
Should we make this nullable or was it always here and we forgot?

qwertyquest:
@iBotPeaches [OpenAI's API reference](https://platform.openai.com/docs/api-reference/responses/create#responses-create-input-input-item-list-item-output-message-content-output-text-annotations-file-citation-filename) marks `filename` as required when passing the `file_citation` annotation back to them as an input item (actually how I discovered this).

image

Looks their [official Node SDK](https://github.com/openai/openai-node/blob/master/src/resources/responses/responses.ts#L3866) marks it as non-nullable too.

iBotPeaches:
Okay, if you can update a test/assertion so we have coverage, then its good to go

qwertyquest:
> Okay, if you can update a test/assertion so we have coverage, then its good to go

Should be good to go!

iBotPeaches:
thanks!

PR #699, opened 2025-10-07
feat: Support string|array output in FunctionToolCallOutput

gembancud:
- Update FunctionToolCallOutput to accept string|array for output parameter
  - This enables returning structured file/image references from function calls
  - Aligns with OpenAI's recent API update for file outputs in tool results
  - No breaking changes: existing string outputs continue to work

### What:

- [x] New Feature

### Description:

[twitter](https://x.com/OpenAIDevs/status/1971618905941856495)
Openai released support for image and file inputs inside tool call results.
This PR aligns with that


gembancud:
apologies for not including it but the shape should be documented in the openai api reference [here](https://platform.openai.com/docs/api-reference/responses/create#responses-create-input-input-item-list-item-function-tool-call-output-output)

kmk142789 (approved):

PR #704, opened 2025-10-15
feat: Add containers interface to client

knash94:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Just a trivial change, adding the containers method to the ClientContract as it's not immediately obvious that this library supports containers


iBotPeaches:
You are right - I wonder how I missed that in https://github.com/openai-php/client/pull/636/files.

I now wonder how I can prevent this in future. Like a test that confirms all client methods are indeed part of the contract.

iBotPeaches (approved):

iBotPeaches:
thanks!

knash94:
> You are right - I wonder how I missed that in https://github.com/openai-php/client/pull/636/files.
> 
> I now wonder how I can prevent this in future. Like a test that confirms all client methods are indeed part of the contract.

That's a good idea, I'll try thinking of a way to achieve it without being too convoluted for a test 🤔

PR #706, opened 2025-10-22
fix(OpenAI): Handle diarized json response in TranscriptionResponseSegment

olivier-drieux:
### What:

- [ ] Bug Fix

### Description:

TranscriptionResponse now parses both [verbose JSON](https://platform.openai.com/docs/api-reference/audio/verbose-json-object) and [diarized JSON](https://platform.openai.com/docs/api-reference/audio/diarized-json-object) responses. 
Segment fields differ between the two response types. 
This commit updates TranscriptionResponseSegment to handle both responses. 
Note: Fields that differ between responses are now nullable, also, the id property can now be a integer or a string.

### DIFF
--- a/src/Responses/Audio/TranscriptionResponse.php
+++ b/src/Responses/Audio/TranscriptionResponse.php
@@ -12,12 +12,12 @@ use OpenAI\Responses\Meta\MetaInformation;
 use OpenAI\Testing\Responses\Concerns\Fakeable;
 
 /**
- * @implements ResponseContract<array{task: ?string, language: ?string, duration: ?float, segments: array<int, array{id: int, seek: int, start: float, end: float, text: string, tokens: array<int, int>, temperature: float, avg_logprob: float, compression_ratio: float, no_speech_prob: float, transient?: bool}>, words: array<int, array{word: string, start: float, end: float}>, text: string}>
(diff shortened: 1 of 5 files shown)


iBotPeaches:
ah interesting. I have to read up on this - I missed it. Enabled CI to run for now

iBotPeaches:
Thanks - looks good. I want to run it through some real testing when I have some time. I'll adapt one of my samples here and run it through on both ways to be sure - https://github.com/iBotPeaches/openai-php-laravel-test/tree/master/app/Console/Commands

iBotPeaches:
Its odd the docs and your samples have non-null values, but my raw tests return a lot more nulls than I expect. Below is a real call sample

```json
{
	"text": "The quick brown fox jumps over the lazy dog.",
	"segments": [
		{
			"type": "transcript.text.segment",
			"text": " The quick brown fox jumps over the lazy dog.",
			"speaker": "A",
			"start": 0,
			"end": 2.1499999999999995,
			"id": "seg_0"
		}
	],
	"usage": {
		"type": "tokens",
		"total_tokens": 172,
		"input_tokens": 51,
		"input_token_details": {
			"text_tokens": 11,
			"audio_tokens": 40
		},
		"output_tokens": 121
	}
}
```

iBotPeaches:
thanks!

PR #711, opened 2025-11-06
feat: Responses streaming types

madskildegaard:
### What:

- [ ] Bug Fix
- [X] New Feature

### Description:

This pull request adds `type` to every streamed object in the Responses endpoint, as seen in the [OpenAI API Reference](https://platform.openai.com/docs/api-reference/responses-streaming/response).

The API Reference also shows that **response.created**, **response.in_progress**, **response.completed**, **response.failed** and **response.incomplete** has a `type`, `response` and `sequence number`. Currently only the `response` object is returned, which is made by the `CreateResponse` class. I've made it so that they now return a `Response` class, which has `type: string`, `response: CreateResponseType` and `sequence_number: int`.

This is a pretty big change, but I believe it is necessary in order to conform to the OpenAI spec more closely.


iBotPeaches:
Is there something in specific you hit that required this? Since we switch off the type to build the sub streaming set, we technically know it. I'm guessing maybe the toArray() not returning it was a bug?

The missing types on response is a good call out. Some types we made fresh streaming classes for, others seemed like they reuse the main response class.

iBotPeaches (approved):
Ran it through my Response tests and all good. Confirmed the changes to Response makes sense. Don't love `response->response`, but I can't think of a better name.

Slight breaking change here, but since we aren't 1.0.0 yet - I believe thats alright.

madskildegaard:
> I'm guessing maybe the toArray() not returning it was a bug?

Exactly. I was running some tests using the client, that required the items to be the same as the OpenAI API Reference says, which is why I recognized the difference.


> Don't love `response->response`, but I can't think of a better name.

I agree - don't like the naming either. Also, using the class `CreateResponse` inside the class `Response` also might be a bit confusing, but like you I couldn't come up with a better name for either of these things.

PR #713, opened 2025-11-11
Response Retrieve streamed

boboldehampsink:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Adds ability to resume a background stream from a response id and cursor:

```php
$stream = OpenAI::responses()->retrieveStreamed($rid, [
  'starting_after' => $cursor,
]);
```

### Related:

https://platform.openai.com/docs/guides/background#streaming-a-background-response

### DIFF
--- a/tests/Resources/Responses.php
+++ b/tests/Resources/Responses.php
@@ -323,6 +323,7 @@ test('retrieve streamed', function () {
     );
 
     $client = mockStreamClient('GET', 'responses/resp_67ccf18ef5fc8190b16dbee19bc54e5f087bb177ab789d5c', [
+        'starting_after' => '2',
         'stream' => 'true',
     ], $response);
 


iBotPeaches:
With a new method like this - I don't exactly remember it all by heart but you'll need to touch the interface, the mock and a simple test to keep the coverage.

boboldehampsink:
Yes this is more of a working draft now. To have discussion over I guess.

boboldehampsink:
@iBotPeaches improved code, added interface methods, docs and tests.

iBotPeaches:
Do you have any further context on the string `true`? I read the docs and it seems its a regular `bool` like the other stream endpoints. Rest looks solid - just trying to understand the string/true.

boboldehampsink:
> Do you have any further context on the string `true`? I read the docs and it seems its a regular `bool` like the other stream endpoints. Rest looks solid - just trying to understand the string/true.

When we use the boolean it is being used as `?stream=1` instead of `?stream=true`. This is because retrieve is a GET request. Apparently the OpenAI API wants this to be "true" instead of "1". It fails when passing 1 with an error message (out of the top of my head) of "cannot parse stream as boolean"

boboldehampsink:
We could also adjust the Payload::toRequest method to handle this, but I'm not sure if it applies to every endpoint.

boboldehampsink:
Also see https://www.php.net/manual/en/function.http-build-query.php#122232

iBotPeaches:
Ahh interesting. Thanks for all the info. Let me mess with this a bit (see impact on other endpoints, etc) and see.

iBotPeaches:
thanks!

Issue #721, opened 2025-12-10
[Feature]: Add webhook utilities as groundwork for Framework-specific integrations


Radiergummi:
This is a sister issue to openai-php/laravel#186.

To add support for webhooks in the Laravel (and possibly also Symfony?) package(s), two utilities should be added to the client package:

1. A [Standard Webhooks](https://www.standardwebhooks.com) compatible signature verifier* to cryptographically validate the signature on OpenAI webhook request, and
2. An Enum of [webhook events](https://platform.openai.com/docs/api-reference/webhook-events).

Duplication into the Framework-specific packages wouldn't make sense, IMH, and they may also be useful to users using the client package on its own.

Regarding the signature verifier, it must be noted that there is an existing package provided by the Standard Webhooks authors available already, but it [hasn't ever been published on Packagist in a stable version](https://github.com/standard-webhooks/standard-webhooks/issues/249), and comes with a few other downsides. I'd rather re-implement the ~90 lines than introduce an unstable dependency.

PR #722, opened 2025-12-10
Feat: Add Webhook Signature Verification

Radiergummi:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

This PR adds
1. a new enum for all webhook events [as listed in the OpenAI docs](https://platform.openai.com/docs/api-reference/webhook-events), and
2. a signature verifier to validate webhook signature headers OpenAI sends with webhook requests as described in [their documentation](https://platform.openai.com/docs/guides/webhooks#verifying-webhook-signatures) and the [Standard Webhooks Spec](https://github.com/standard-webhooks/standard-webhooks/blob/main/spec/standard-webhooks.md#verifying-webhook-authenticity).

This provides some primitives required to add webhook support to their app, which allows for asynchronous LLM workflows where you don't want to (or can't) wait for several minutes until the model provides the response, and instead get notified when the results are available.  
While HTTP request handling and result processing is firmly out of scope for this library itself, the verification mechanism is security sensitive and interested users will definitely benefit from not having to implement it themselves.  
Additionally, it will allow for adding more integrated webhook handling to the framework-specific packages later on.

### Related:

This will solve #721 and provide the requirements for openai-php/laravel#186.

fixes: #721

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ If you or your business relies on this package, it's important to support the de
 - [Meta Information](#meta-information)
 - [Troubleshooting](#troubleshooting)
 - [Testing](#testing)
+- [Webhooks][#webhooks]
 - [Services](#services)
   - [Azure](#azure)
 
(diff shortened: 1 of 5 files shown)


iBotPeaches:
thanks! Looks nice on first peek. Let me setup my side to actually run this through with ngrok and test it out.

PR #728, opened 2026-01-08
Responses\CreateResponseType optional values are not handled properly

bbrala:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

When using LiteLLM to proxy requests i got a few errors while building the request. It seems a few of the nullable attributes are not handled properly. This means if they are missing the package crashed.

I'm fixing this in 2 commits. First off, adjust the array shape definition to make sure the properties are optoinal, this surfaces the errors in the `from` method. Which i will then properly handle.


iBotPeaches (commented):
Could you apply the same patch to the Retrieve method? Its an odd amount of duplication that I never cleaned up 

https://github.com/openai-php/client/blob/main/src/Responses/Responses/RetrieveResponse.php

bbrala:
I've consolidated the two classes, i think there was another error where a specific tool was missing.

PR #737, opened 2026-01-28
feat: add reasoning_content field to CreateResponseMessage 

Kemel91:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Add support for the reasoning_content field in chat completions API responses. This field is added at the same level as content in both streaming and non-streaming requests. The field is marked as optional to handle cases where it's not present in the API response.

### Related:

fixes https://github.com/openai-php/client/issues/517

### DIFF
--- a/tests/Fixtures/Chat.php
+++ b/tests/Fixtures/Chat.php
@@ -738,7 +738,7 @@ function chatCompletionReasoningContent(): array
                 'message' => [
                     'role' => 'assistant',
                     'content' => null,
-                    'reasoning_content' => "Hello world",
+                    'reasoning_content' => 'Hello world',
                 ],
                 'logprobs' => null,
                 'finish_reason' => 'stop',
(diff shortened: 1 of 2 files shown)


iBotPeaches:
Am I blind? I can't find that property on the API reference: https://platform.openai.com/docs/api-reference/chat/object

Kemel91:
> Am I blind? I can't find that property on the API reference: https://platform.openai.com/docs/api-reference/chat/object

The reasoning_content field is indeed not yet listed in the official OpenAI API reference.

However, this field is currently being returned by the API when using certain experimental or yet-to-be-documented features (e.g., when the reasoning option is enabled in the request).

We added support for it in the library to:

Avoid data loss: If the API does return this field, the library will now parse it correctly.

Stay ahead of future updates: New fields often appear in the API before they are officially documented.

Support users already experimenting with the feature: Some developers are already using unreleased capabilities, and the library shouldn’t break or ignore the data.

This is a common practice in SDKs to maintain compatibility with potential API changes before they are formally announced.

iBotPeaches:
Okay, update/create some tests and I think we are good to go.

Cyrille37:
Thanks, guys! 💌

PR #738, opened 2026-02-05
Feature: add support for reasoning content in streamed completions API

Tobiasmidskards:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:
Add support for the reasoning_content field in chat completions API responses for streamed responses. This field is added at the same level as content in both streaming and non-streaming requests. The field is marked as optional to handle cases where it's not present in the API response.

### Related:
Same as but for streaming https://github.com/openai-php/client/pull/737


L3o-pold:
`reasoningContent` is missing from the `toArray` functions

PR #739, opened 2026-02-08
Include response body in RateLimitException message

limonte:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

Before (no clue about why RateLimitException is thrown)

```
RateLimitException   
  Request rate limit has been exceeded.
```

After (with details)

```
RateLimitException   
  Request rate limit has been exceeded: {
  "error": {
    "message": "You exceeded your current quota, please check your plan and billing details. For more information on this error, read the docs: https://platform.openai.com/docs/guides/error-codes/api-errors.",
    "type": "insufficient_quota",
    "param": null,
    "code": "insufficient_quota"
  }
}
```

### DIFF
--- a/src/Exceptions/RateLimitException.php
+++ b/src/Exceptions/RateLimitException.php
@@ -11,6 +11,6 @@ final class RateLimitException extends Exception
 {
     public function __construct(public ResponseInterface $response)
     {
-        parent::__construct('Request rate limit has been exceeded: ' . $response->getBody());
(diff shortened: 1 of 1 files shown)


iBotPeaches:
I think we should expand the class to to hold the response for people catching the exception. Shoving raw HTTP response into the message of exception seems like that could mess with folks error tracking and grouping and might not be robust enough against like a null (204) or other return.

limonte:
filed the issue instead: https://github.com/openai-php/client/issues/743

PR #741, opened 2026-02-13
Fix syntax error in README.md example code

simonschaufi:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Fix example code for testing streams

### DIFF
--- a/README.md
+++ b/README.md
@@ -3057,18 +3057,18 @@ use OpenAI\Testing\ClientFake;
 use OpenAI\Responses\Chat\CreateStreamedResponse;
 
 $client = new ClientFake([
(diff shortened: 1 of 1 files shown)


iBotPeaches:
thanks!

simonschaufi:
Welcome! Can you please also add an example how the content of the file.txt can look like?

iBotPeaches:
Have you tried some of the sample fixtures we have for streams?

https://github.com/openai-php/client/tree/main/tests/Fixtures/Streams

simonschaufi:
I used this example here https://github.com/openai-php/client/blob/main/tests/Fixtures/Streams/ChatCompletionCreate.txt but I get an `Missing event type in streamed response` exception.

PR #744, opened 2026-02-25
fix: default tool call type to function for Mistral compatibility

### LINKED ISSUE #397: [Bug]: MistralAI- Undefined array key "type"
### Description

Hi,

I'm using this API implementation with both OpenAI models and MistralAI ones.

My implementation works perfectly with both, except for the tool_calls function which is the only thing that goes wrong with MistralAI models.

When calling a function with a compliant model (example: open-mixtral-8x22b), I get this error message:

```
Message: Undefined array key "type"

Filename: Chat/CreateResponseToolCall.php

Line Number: 23
```

It only happens when the model intends to use the tool.
It's able to recognise its tools though, but as soon as it tries to use them, there is this error occuring.

### Steps To Reproduce

Create a MistralAI account and get an API key.

Use this base URI: api.mistral.ai/v1

Make a chat resource with a tool and use it with a MistralAI compliant model (example: open-mixtral-8x22b).

### OpenAI PHP Client Version

v0.8.5

### PHP Version

8.3

### Notes

_No response_

ivanmestre:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

The Mistral API currently omits the `type` attribute in its tool call responses, which causes an `Undefined array key "type"` exception. This commit defaults the missing `type` to `'function'` (the standard supported type per OpenAI documentation) to prevent runtime errors and ensure broader compatibility with OpenAI-compatible APIs.

### Related:

Fixes #397

### DIFF
--- a/src/Responses/Chat/CreateResponseToolCall.php
+++ b/src/Responses/Chat/CreateResponseToolCall.php
@@ -13,13 +13,13 @@ final class CreateResponseToolCall
     ) {}
 
     /**
-     * @param  array{id: string, type: string, function: array{name: string, arguments: string}}  $attributes
+     * @param  array{id: string, type?: string, function: array{name: string, arguments: string}}  $attributes
      */
     public static function from(array $attributes): self
     {
         return new self(
             $attributes['id'],
-            $attributes['type'],
+            $attributes['type'] ?? 'function',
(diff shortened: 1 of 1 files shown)


iBotPeaches (approved):
Debated in my head - fine with defaulting this to `function`

PR #747, opened 2026-03-10
fix: vector store description

madskildegaard:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

This PR adds `description` to the **VectorStore** object which is expected as seen [here](https://developers.openai.com/api/reference/resources/vector_stores/methods/create) and [here](https://developers.openai.com/api/reference/resources/vector_stores/methods/list).


iBotPeaches:
thanks!

PR #748, opened 2026-03-10
fix: vector store file contract missing updateAttributes function

madskildegaard:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

This PR adds the missing `updateAttributes` function to the `VectorStoresFilesContract`. The function is already implemented in [VectorStoresFiles](https://github.com/openai-php/client/blob/main/src/Resources/VectorStoresFiles.php), but the contract missing it causes `$client->vectorStores()->files()->updateAttributes()` to complain about "Undefined method 'updateAttributes'." in linters.


iBotPeaches (approved):

iBotPeaches:
thanks again!

PR #751, opened 2026-03-27
fix: reasoning streaming

L3o-pold:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Using streaming response, reasoning was not handled.

### DIFF
--- a/src/Responses/Chat/CreateStreamedResponseDelta.php
+++ b/src/Responses/Chat/CreateStreamedResponseDelta.php
@@ -12,12 +12,13 @@ final class CreateStreamedResponseDelta
     private function __construct(
         public readonly ?string $role,
         public readonly ?string $content,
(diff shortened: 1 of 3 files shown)


iBotPeaches (changes requested):
I don't see it here - https://developers.openai.com/api/reference/resources/chat/subresources/completions/streaming-events

Are docs misleading?

L3o-pold:
@iBotPeaches I think it's more for OpenAI compilant server like [VLLM](https://docs.vllm.ai/en/stable/examples/online_serving/openai_chat_completion_with_reasoning_streaming/)

iBotPeaches (approved):
I don't love growing these for non-openai, but Chat is pretty much dead for pure OpenAI usages as I'm guessing most moved to Responses. So I think its fine to expand Chat as long as all properties are nullable like done here.

PR #756, opened 2026-04-07
fix VectorStoreResponseFileCounts attribute order

madskildegaard:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

`failed` and `cancelled` counts were swapped in the response.


iBotPeaches:
hmm that sucks. I wonder why none of our assertions caught this. Nice find

PR #763, opened 2026-04-22
feat: add Astraflow provider support

ucloudnb666:
## Add Astraflow provider support

[Astraflow](https://astraflow.ucloud.cn/) (by UCloud / 优刻得) is an OpenAI-compatible AI model aggregation platform that supports 200+ models via two regional endpoints:

| Region | Base URI | Env var |
|--------|----------|---------|
| Global | `https://api-us-ca.umodelverse.ai/v1` | `ASTRAFLOW_API_KEY` |
| China  | `https://api.modelverse.cn/v1`        | `ASTRAFLOW_CN_API_KEY` |

### Changes

| File | What changed |
|------|-------------|
| `src/Astraflow.php` | New global-namespace facade — mirrors `src/OpenAI.php`; exposes `client()` (global endpoint), `clientCn()` (China endpoint), and `factory()` |
| `composer.json` | Registers `src/Astraflow.php` under `autoload.files` so the class is available without an explicit `use` |
| `tests/Astraflow.php` | Unit tests that mirror `tests/OpenAI.php`, covering both endpoints and the full factory API |
| `README.md` | Documents the new **Astraflow** service under the *Services* section (Table of Contents + prose) |

### Usage

```php
// Global endpoint
$client = Astraflow::client(getenv('ASTRAFLOW_API_KEY'));

// China endpoint
$client = Astraflow::clientCn(getenv('ASTRAFLOW_CN_API_KEY'));

// Fine-grained factory — pick any endpoint / HTTP client / headers
$client = Astraflow::factory()
    ->withApiKey(getenv('ASTRAFLOW_API_KEY'))
    ->withBaseUri(Astraflow::BASE_URI)   // or Astraflow::BASE_URI_CN
    ->make();
```

Sign up for an API key at **https://astraflow.ucloud.cn/**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ If you or your business relies on this package, it's important to support the de
 - [Testing](#testing)
 - [Webhooks][#webhooks]
 - [Services](#services)
+  - [Astraflow](#astraflow)
   - [Azure](#azure)
 
 ## Get Started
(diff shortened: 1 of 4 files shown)


iBotPeaches (changes requested):
This would be the first time we introduced a new client for a specific provider, which would suggest to population that it is officially supported and 1st party.

We only target OpenAI and providers that are drop-in replacements (mainly originally built for the OpenAI on Azure) are what that was intended for. If this works for the Gemini OpenAI compatible service or this - great, but I don't think we are going to make specific clients for it.

At this time I would prefer this remains as a configuration of the Client without adding it this way.

Thanks for the PR and understanding.

PR #766, opened 2026-05-05
Provide a clearer JsonException message when HttpTransporter receives invalid response.

arbazargani:
Fix handling of invalid response.

### What:

* [x] Bug Fix
* [ ] New Feature

### Description:

This PR improves the error handling in `HttpTransporter` when an invalid or malformed response is returned by the service provider.

Previously, such cases could result in unclear or misleading exceptions. In scenarios where the service provider returned a non-JSON response (such as an HTML error page like `403 Forbidden` from nginx, or a CDN-blocked request), the client would typically throw a generic `syntax error`, which made it difficult to understand the root cause.

This change ensures that a proper `JsonException` is thrown with a more descriptive and accurate error message, making debugging easier and behavior more predictable. It explicitly handles cases where the response body is not valid JSON and provides clearer insight into what actually went wrong.

### Related:

N/A

### DIFF
--- a/src/Exceptions/UnserializableResponse.php
+++ b/src/Exceptions/UnserializableResponse.php
@@ -12,6 +12,6 @@ final class UnserializableResponse extends Exception
 {
     public function __construct(JsonException $exception, public ResponseInterface $response)
     {
-        parent::__construct($exception->getMessage(), 0, $exception);
+        parent::__construct($exception->getMessage() . ' or the server returned an invalid response.', 0, $exception);
     }
 }


iBotPeaches:
At the moment not really feeling the use for this. The `JsonException` is clear that we got an issue. Other classes tend to catch different 5xx and 4xx errors. This just adds noise and confusion to an exception.

iBotPeaches:
Thanks, but decided against this. It just adds clutter to an existing message.

PR #718, opened 2025-11-27
add support for extra_content in chat response tool calls returned by Gemini models

ezequiel-safirsztein-momentum:
### What:

- [ ] Bug Fix
- [X] New Feature

### Description:

This PR introduces support for the ⁠extra_content field, which is returned by certain Gemini models (specifically those utilizing the reasoning signature) during tool calls.
When interacting with these models, the API provides ⁠extra_content alongside tool calls. To maintain the integrity of the conversation and satisfy the model's reasoning signature requirements, this content must be preserved and passed back to the API when submitting tool outputs.
Changes:
	• Updated the response handling to parse and expose ⁠extra_content from tool call responses. 	
         • Updated the payload builder to include ⁠extra_content when submitting tool outputs back to the model.

PR #768, opened 2026-05-15
add support for extra_content in chat response tool calls returned by Gemini models

leonardfischer:
### What:

- [X] Bug Fix
- [ ] New Feature

### Description:

This is a newer fix, based on #718 in order to fix the Gemini reponse. Gemini sends 'extra_content' when using tools in order to transfer the `thought_signature`. This is how it looks in the raw response:

```json
{
  "choices": [
    {
      "finish_reason": "tool_calls",
      "index": 0,
      "message": {
        "role": "assistant",
        "tool_calls": [
          {
            "extra_content": {
              "google": {
                "thought_signature": "<token>"
              }
            },
            "function": {
              "arguments": "{}",
              "name": "function-name"
            },
            "id": "<id>",
            "type": "function"
          },
          // ...
        ]
      }
    }
  ],
  // ...
```

### Related:

See other PR #718 - this one should take care of failed checks

closes: #718

### DIFF
--- a/src/Responses/Chat/CreateResponseToolCall.php
+++ b/src/Responses/Chat/CreateResponseToolCall.php
@@ -10,10 +10,11 @@ final class CreateResponseToolCall
         public readonly string $id,
         public readonly string $type,
         public readonly CreateResponseToolCallFunction $function,
+        public readonly array $extraContent,
     ) {}
 
     /**
-     * @param  array{id: string, type?: string, function: array{name: string, arguments: string}}  $attributes
+     * @param  array{id: string, type?: string, function: array{name: string, arguments: string}, extra_content: array{name: string, arguments: string}|null}  $attributes
      */
     public static function from(array $attributes): self
     {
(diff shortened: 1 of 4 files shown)


iBotPeaches:
You'll need to type the array shape of this since we are in highest PHPStan mode. Also could we get some tests to confirm OpenAI who doesn't send this doesn't have it returned? Then likewise a test for when its included that it comes back in `toArray` and other functions.

leonardfischer:
Hey @iBotPeaches alright - I'm on it :)

leonardfischer:
Sooo... After quite some back- and forth I managed to satisfy all linters and type-checkers (sorry, I noticed to late that this stuff can be checked locally - I'm more used to githooks that do this before committing).

I also added one more test and verified in the existing test that the new "extra_content" would not appear

leonardfischer:
Hey @iBotPeaches could you (re-) check the current state? Should I improve something more :thinking: ?

iBotPeaches (approved):
I don't love the expanse because its for another model, but on the flip side. Most folks using pure OpenAI have moved onto to Response API, so to add something that no one will notice unless using Gemini seems fine.

PR #775, opened 2026-06-04
Document environment-based custom base URI

cerebrixos:
This keeps the default OpenAI behavior unchanged while documenting a portable way to configure an OpenAI-compatible endpoint through an environment variable.\n\nWhy this may be useful:\n- local inference servers\n- private model gateways\n- self-hosted routers\n- observability and enterprise control planes\n\nThe example uses the existing withBaseUri() API and does not add a provider-specific integration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,17 @@ $client = OpenAI::factory()
     ->make();
 ```
 
+For applications that need to switch between OpenAI and another
+OpenAI-compatible endpoint in different environments, the base URI can be read
+from an environment variable:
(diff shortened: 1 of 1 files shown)


iBotPeaches (changes requested):
Thanks, but I'm not seeing the need for this. The section above literally has all the configuration, including the `withApiKey` and `withBaseUri`. We don't want to call out how to use envs, because every system will do it differently depending on plain php, Laravel or Symfony.

PR #786, opened 2026-07-02
fix(OpenAI): default missing optional keys to null in from()

### LINKED ISSUE #781: Unchecked access to optional keys in $attributes in public static function from() in multiple classes.
I am using the library not only to talk to open ai, but also to talk to azure ai agents that fully support the responses api. It looks like azure ai skips empty attributes in their response and openai does not. This causes problems in some classes where the code tries to access the array index of optional attribute without checking if it exists:
  
  
1) Example - WebSearchUserLocation.php:

The constructor shows that "city" is optional:

`public readonly ?string $city`

But in public static function from() the acces to it is not checked:

`city: $attributes['city'],`

Much better would be this:

`city: $attributes['city'] ?? null,`
  
  
2) Example - OutputMcpCall.php:

The construtor shows that "approvalRequestId" is optional:

`public readonly ?string $approvalRequestId = null,`

But the same in public static function from(): 

`approvalRequestId: $attributes['approval_request_id'],`

While this would be much better:

`approvalRequestId: $attributes['approval_request_id'] ?? null,`
  
  
I am not sure if you are generating this could automatically. If not I could start making pull request for all those cases where I run into a problem. Should I do that?

lazerg:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

`WebSearchUserLocation::from()` and `OutputMcpCall::from()` read their nullable keys straight from `$attributes`, so a response that omits those optional keys triggers an "Undefined array key" warning. This defaults them to `null` with `?? null`, the same way the other `from()` methods already handle optional keys.

### Related:

fixes: #781

### DIFF
--- a/src/Responses/Responses/Output/OutputMcpCall.php
+++ b/src/Responses/Responses/Output/OutputMcpCall.php
@@ -69,9 +69,9 @@ final class OutputMcpCall implements ResponseContract
             type: $attributes['type'],
             arguments: $attributes['arguments'],
             name: $attributes['name'],
-            approvalRequestId: $attributes['approval_request_id'],
+            approvalRequestId: $attributes['approval_request_id'] ?? null,
             error: $errorType,
-            output: $attributes['output'],
+            output: $attributes['output'] ?? null,
         );
     }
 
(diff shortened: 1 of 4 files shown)


iBotPeaches:
Any further context on the `set_error_handler` usage? None of our other tests do that, so curious what you hit or trying to test with that.

lazerg:
Undefined array key only warns in PHP, doesn't throw. Value's null either way, so a plain assertion wouldn't catch the regression without escalating it. Scoped to that one call via try/finally so it doesn't leak into other tests.

iBotPeaches (approved):

PR #787, opened 2026-07-06
fix(OpenAI): Support streaming 'compaction' output

e4se:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

- Add compaction output parsing to Responses streaming OutputItem
- Add a stream fixture and regression test for response.output_item.done with a compaction item

## Why
Streaming response output items could not parse the new compaction output object even though non-streamed Responses output already supported it.


iBotPeaches (approved):
Looks great. Thanks for catching my mistake, I forgot to do the streaming aspect.

PR #788, opened 2026-07-20
fix(OpenAI): Preserve function call namespaces

e4se:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

Preserve the optional namespace returned for namespaced Responses API function calls. Regular function calls without a namespace remain supported and expose a null namespace value.

This also adds coverage for parsing, array access, serialization, and responses without the optional namespace field.


iBotPeaches (approved):

iBotPeaches:
thanks!

PR #793, opened 2026-07-28
docs: document MiniMax OpenAI-compatible service and regional endpoints

octo-patch:
Reason: Document the MiniMax OpenAI-compatible service, including its two regional endpoints and current text models.

## Changes

- Add a `### MiniMax` section under `## Services` in the README, mirroring the existing Azure service documentation.
- Document the two regional OpenAI-compatible base URIs (`api.minimax.io/v1` for `global_en` and `api.minimaxi.com/v1` for `cn_zh`).
- Show how to configure the existing factory (`withBaseUri`) to target MiniMax and how to select the current text models (`MiniMax-M3`, `MiniMax-M2.7`) through the request `model` parameter.
- Add the section to the table of contents.

## How it follows the existing pattern

The Azure service is already documented as a README `Services` section that composes the generic factory extension points (`src/Factory.php` `withBaseUri`/`withHttpHeader`/`withQueryParam` and `src/ValueObjects/Transporter/BaseUri.php`). MiniMax is OpenAI-compatible and is added the same way, without introducing provider-specific source code, which keeps the client provider-agnostic.

## Checks

- Verified the new `### MiniMax` heading produces the `#minimax` anchor referenced in the table of contents.
- This change is documentation-only; no PHP source files were modified, so the PHP lint/type/test toolchain is unaffected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ If you or your business relies on this package, it's important to support the de
 - [Webhooks][#webhooks]
 - [Services](#services)
   - [Azure](#azure)
+  - [MiniMax](#minimax)
 
 ## Get Started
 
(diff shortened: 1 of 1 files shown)


iBotPeaches:
Thanks, but the goal is to keep that section for official OpenAI compatible services - which at this time is only Azure.

PR #795, opened 2026-08-11
fix(OpenAI): optimize streamed response processing

e4se:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

`StreamResponse` previously read SSE bodies one byte at a time. Large streamed payloads such as `encrypted_content` could therefore trigger millions of `read(1)` calls, while `trim()` and generator-local references kept extra copies of the raw JSON in memory.

This change reads the stream in 64 KiB chunks and preserves unread data in an internal line buffer. It avoids trimming large JSON values, releases the raw payload before yielding the DTO, and retries temporary empty reads until EOF. The response schema is unchanged and `encrypted_content` is still returned normally.

A 130 MiB `response.output_item.done` compaction event containing `encrypted_content` shows the impact: the previous byte-at-a-time implementation would require roughly 136 million reads, while the updated reader performs 2,081 reads. The optimized path processed the event in 0.2058 seconds with an additional peak memory usage of 272,662,528 bytes (about 260 MiB).

A direct before-and-after benchmark with an 8 MiB payload measured:

- processing time: 1.7393 seconds → 0.0180 seconds (about 96.6× faster)
- read calls: 8,388,794 → 129 (about 65,000× fewer)
- additional peak memory: 25,214,976 → 16,809,984 bytes (about 33% lower)

Regression tests cover large buffered events and an empty read before EOF.


iBotPeaches (approved):

iBotPeaches:
thanks!

PR #798, opened 2026-08-18
fix(OpenAI): allow null ranking_options and max_num_results in FileSearchTool

### LINKED ISSUE #796: [Bug]: Responses FileSearchTool rejects optional null ranking_options
## Description

`OpenAI\Responses\Responses\Tool\FileSearchTool::from()` cannot hydrate a schema-valid Responses API `file_search` tool when `ranking_options` is omitted or returned as `null`.

The official OpenAI response model declares `ranking_options` as optional (`Optional[RankingOptions] = None`):
https://github.com/openai/openai-python/blob/main/src/openai/types/responses/file_search_tool.py

However, `FileSearchTool` declares `rankingOptions` as a non-nullable `FileSearchRankingOption` and unconditionally executes:

```php
FileSearchRankingOption::from($attributes['ranking_options'])
```

This prevents callers from consuming an otherwise successful Responses API result.

Expected behaviour: `FileSearchTool::from()` accepts an omitted or null `ranking_options` field and represents it as nullable without inventing ranking values. Explicit ranking options should continue to hydrate unchanged.

Actual behaviour:

```text
OpenAI\Responses\Responses\Tool\FileSearchRankingOption::from(): Argument #1 ($attributes) must be of type array, null given
```

## Steps To Reproduce

Install the latest release and execute:

```php
<?php

require 'vendor/autoload.php';

OpenAI\Responses\Responses\Tool\FileSearchTool::from([
    'type' => 'file_search',
    'vector_store_ids' => ['vs_example'],
    'filters' => null,
    'max_num_results' => 10,
    'ranking_options' => null,
]);
```

The call throws from `FileSearchTool.php:60` when null is passed to `FileSearchRankingOption::from()`.

Removing the `ranking_options` key also fails because the hydrator accesses it unconditionally.

## OpenAI PHP Client Version

v0.20.1

## PHP Version

8.3.6

## Notes

- v0.20.1 is the latest release as of 18 August 2026.
- `max_num_results` is deliberately populated in the reproduction so this report isolates `ranking_options` handling.
- OpenAI's documented defaults are `ranker: auto` and `score_threshold: 0`, but a response client should preserve the schema's optional field rather than manufacture those values during hydration.

### LINKED ISSUE #797: [Bug]: Responses FileSearchTool rejects optional null max_num_results
## Description

`OpenAI\Responses\Responses\Tool\FileSearchTool::from()` cannot hydrate a schema-valid Responses API `file_search` tool when `max_num_results` is omitted or returned as `null`.

The official OpenAI response model declares `max_num_results` as optional (`Optional[int] = None`):
https://github.com/openai/openai-python/blob/main/src/openai/types/responses/file_search_tool.py

However, `FileSearchTool` declares `maxNumResults` as a non-nullable `int` and unconditionally executes:

```php
maxNumResults: $attributes['max_num_results'],
```

This prevents callers from consuming an otherwise successful Responses API result.

Expected behaviour: `FileSearchTool::from()` accepts an omitted or null `max_num_results` field and represents it as `?int` without inventing a result cap. Explicit integer values should continue to hydrate unchanged.

Actual behaviour:

```text
OpenAI\Responses\Responses\Tool\FileSearchTool::__construct(): Argument #4 ($maxNumResults) must be of type int, null given
```

Related: this is the same `FileSearchTool` hydrator as [#796](https://github.com/openai-php/client/issues/796), which covers optional `ranking_options`. That report deliberately populated `max_num_results` so the two optional fields could be tracked separately.

## Steps To Reproduce

Install the latest release and execute:

```php
<?php

require 'vendor/autoload.php';

OpenAI\Responses\Responses\Tool\FileSearchTool::from([
    'type' => 'file_search',
    'vector_store_ids' => ['vs_example'],
    'filters' => null,
    'max_num_results' => null,
    'ranking_options' => [
        'ranker' => 'auto',
        'score_threshold' => 0.0,
    ],
]);
```

The call throws from `FileSearchTool.php` when `null` is passed to the non-nullable `int $maxNumResults` constructor parameter.

Removing the `max_num_results` key also fails because the hydrator accesses it unconditionally.

## OpenAI PHP Client Version

v0.20.1

## PHP Version

8.3.6

## Notes

- v0.20.1 is the latest release as of 18 August 2026.
- `ranking_options` is deliberately populated in the reproduction so this report isolates `max_num_results` handling.
- A compatible response client should preserve the schema's optional field rather than manufacture a retrieval cap during hydration.

lazerg:
### What:

- [x] Bug Fix
- [ ] New Feature

### Description:

`FileSearchTool::from()` treats `max_num_results` and `ranking_options` as required, but the Responses API returns both as optional (`Optional[int] = None` and `Optional[RankingOptions] = None` upstream). A response that omits or nulls either field currently throws instead of hydrating. Both fields are now nullable and only hydrated when present.

### Related:

fixes: #796
fixes: #797

### DIFF
--- a/src/Responses/Responses/Tool/FileSearchTool.php
+++ b/src/Responses/Responses/Tool/FileSearchTool.php
@@ -13,7 +13,7 @@ use OpenAI\Testing\Responses\Concerns\Fakeable;
  * @phpstan-import-type ComparisonFilterType from FileSearchComparisonFilter
  * @phpstan-import-type CompoundFilterType from FileSearchCompoundFilter
  *
- * @phpstan-type FileSearchToolType array{type: 'file_search', vector_store_ids: array<int, string>, filters: ComparisonFilterType|CompoundFilterType|null, max_num_results: int, ranking_options: RankingOptionType}
+ * @phpstan-type FileSearchToolType array{type: 'file_search', vector_store_ids: array<int, string>, filters: ComparisonFilterType|CompoundFilterType|null, max_num_results: int|null, ranking_options: RankingOptionType|null}
  *
  * @implements ResponseContract<FileSearchToolType>
  */
@@ -34,8 +34,8 @@ final class FileSearchTool implements ResponseContract
         public readonly string $type,
         public readonly array $vectorStoreIds,
         public readonly FileSearchComparisonFilter|FileSearchCompoundFilter|null $filters,
-        public readonly int $maxNumResults,
-        public readonly FileSearchRankingOption $rankingOptions,
+        public readonly ?int $maxNumResults,
+        public readonly ?FileSearchRankingOption $rankingOptions,
     ) {}
 
     /**
@@ -56,8 +56,8 @@ final class FileSearchTool implements ResponseContract
             type: $attributes['type'],
             vectorStoreIds: $attributes['vector_store_ids'],
             filters: $filters,
-            maxNumResults: $attributes['max_num_results'],
-            rankingOptions: FileSearchRankingOption::from($attributes['ranking_options']),
+            maxNumResults: $attributes['max_num_results'] ?? null,
+            rankingOptions: isset($attributes['ranking_options']) ? FileSearchRankingOption::from($attributes['ranking_options']) : null,
         );
     }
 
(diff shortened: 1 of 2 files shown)


iBotPeaches (approved):

iBotPeaches:
thanks! perfect commit style, perfect test and formatting and everything passed.

